PR #29, opened 2020-10-11
Add pino

DeltaEvo:


src/index.js
```diff
@@ -51,6 +55,10 @@ function handle_login(world) {
       experience: 0,
     }
 
+    client.on('error', (error) => {
+      throw error
```
Sceat, line 59: u sure about throwing an unhandled error at runtime ?

Sceat (approved):

src/index.js
DeltaEvo: It's not good for production but for development it help you to see this error since it's not logged anywhere, we should see if we want to change this for production

PR #33, opened 2020-10-28
Simplify code by removing handler list

DeltaEvo:


src/index.js
```diff
@@ -61,17 +49,16 @@ function handle_login({ world, handlers }) {
...
 
     login(state)
     send_resource_pack(state)
     position_change_event(state)
     chunk_change_event(state)
     update_chunks(state)
-    for (const handler of handlers) handler(state)
+    spawn_villager(state)
+    open_trade(state)
+    spawn_mob(state)
```
Sceat, line 61: you could here use the reducer to avoid repeating `state` but feels ok anyway tho

src/trade/trade.js
```diff
@@ -2,60 +2,63 @@ import { version } from '../settings.js'
...
-              inputItem2: {
-                present: true,
-                itemId: diamond,
-                itemCount: 1,
-              },
-              maximumNbTradeUses: 99999,
+export function open_trade({ client, world }) {
+  const { villagers, windowId } = world.trade
+  client.on('use_entity', ({ target, mouse, sneaking }) => {
+    if (villagers.includes(target) && mouse === 2 && sneaking === false) {
```
Sceat, line 28: Magic `2` ? 😏

Sceat (changes requested):

src/trade/trade.js
DeltaEvo: It's from the original code cc @Mathis-Paul-62 I will create an issue for that

Sceat (approved):

src/index.js
DeltaEvo: For now since we not have much handlers I think I will keep this, we will refactor this when we have more functions

PR #35, opened 2020-11-01
basic level implementation

Vahelnir:
Here is a basic implementation of the level system based on the ExpStorage.java.

As you can see, i chose to hardcode the levels (for now, should be moved to a config file or whatever) instead of using the `generate_levels` function. I kept it in the code so that you can choose which way you want to generate/load the levels. 

Not sure if i should have used jsdoc or not, could be removed :shrug:

This is the first time i'm playing with fp so there should be many things to review.

### DIFF
--- a/src/experience.js
+++ b/src/experience.js
@@ -1,0 +1,164 @@
+const levels = [
+  0, // levels starts at 1
+  0,
+  121,
+  715,
+  1650,
+  3080,
+  5280,
+  8030,
+  11550,
+  15950,
+  21120,
+  27720,
+  35860,
+  45100,
+  55550,
+  67100,
+  82500,
+  100100,
+  126500,
+  156200,
+  188100,
+  222200,
+  258500,
+  297000,
+  341000,
+  388300,
+  438350,
+  492800,
+  553300,
+  617100,
+  683760,
+  755700,
+  830500,
+  911900,
+  1001000,
+  1100000,
+  1219000,
+  1358000,
+  1517000,
+  1696000,
+  1895000,
+  2114000,
+  2353000,
+  2612000,
+  2891000,
+  3190000,
+  3509000,
+  3848000,
+  4207000,
+  4586000,
+  4985000,
+  5404000,
+  5843000,
+  6302000,
+  6781000,
+  7280000,
+  7799000,
+  8338000,
+  8897000,
+  9476000,
+  10075000,
+  10694000,
+  11333000,
+  11992000,
+  12671000,
+  13370000,
+  14089000,
+  14828000,
+  15587000,
+  16366000,
+  17165000,
+  17984000,
+  18823000,
+  19682000,
+  20561000,
+  21460000,
+  22379000,
+  23318000,
+  24277000,
+  25256000,
+  26255000,
+  27274000,
+  28313000,
+  29372000,
+  30451000,
+  31550000,
+  32669000,
+  33808000,
+  34967000,
+  36146000,
+  37345000,
+  38564000,
+  39803000,
+  41062000,
+  42341000,
+  43640000,
+  44959000,
+  46298000,
+  47657000,
+  49036000,
+  50435000,
+]
+
+const reversed_levels = [...levels.entries()].reverse()
+
+/**
+ * add some experience to the player
+ * @param {any} state
+ * @param {number} experience
+ */
+export function add_experience(state, experience) {
+  const new_state = { ...state, experience: state.experience + experience }
+  update_experience(new_state)
+  return new_state
+}
+
+/**
+ * find the current level and the remaining
+ * experience from the total experience
+ * @param {number} experience
+ * @returns {{level: number, remaining_experience: number}}
+ */
+export function experience_to_level(total_experience) {
+  const [current_level, current_level_experience] = reversed_levels.find(
+    ([level, level_experience]) => level_experience <= total_experience
+  )
+  return {
+    level: current_level,
+    remaining_experience: total_experience - current_level_experience,
+  }
+}
+
+/**
+ * calculate the level progression from the
+ * current level and the remaining experience
+ * @param {{level: number, remaining_experience: number}} level_data
+ */
+export function level_progress({ level, remaining_experience }) {
+  const current_level_experience = levels[level]
+  if (level + 1 >= levels.length) {
+    return 0
+  }
+  const next_level_experience = levels[level + 1]
+  return (
+    remaining_experience / (next_level_experience - current_level_experience)
+  )
+}
+
+/**
+ * update the player experience from the state
+ * @param {any} state
+ */
+export function update_experience(state) {
+  const { client, experience: total_experience } = state
+  const { level, remaining_experience } = experience_to_level(total_experience)
+  const progress = level_progress({ level, remaining_experience })
+
+  client.write('experience', {
+    totalExperience: total_experience,
+    level,
+    experienceBar: progress,
+  })
+}
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,7 @@ import { register_villagers, spawn_villager } from './trade/spawn_villager.js'
 import { register_trades, open_trade } from './trade/trade.js'
 import { register_mobs, spawn_mob } from './mobs/spawn_mob.js'
 import { send_resource_pack } from './resource_pack.js'
+import { update_experience } from './experience.js'
 
 const server = protocol.createServer({ version, 'online-mode': online_mode })
 
@@ -46,12 +47,14 @@ function handle_login(world) {
       }),
       events: new EventEmitter(),
       gameMode: 2,
+      experience: 0,
     }
 
     // Handle next login
     handle_login({ ...world, lastEntityId: world.lastEntityId + 1 })
 
     login(state)
+    update_experience(state)
     send_resource_pack(state)
     position_change_event(state)
     chunk_change_event(state)


src/index.js
```diff
@@ -46,12 +47,14 @@ function handle_login(world) {
...
       events: new EventEmitter(),
       gameMode: 2,
+      exp: 0,
     }
 
     // Handle next login
     handle_login({ ...world, lastEntityId: world.lastEntityId + 1 })
 
     login(state)
+    player_update_experience(state)
```
Vahelnir, line 57: Not sure if that should be here.
Maybe in the login.js ?

src/player/experience.js
```diff
@@ -0,0 +1,233 @@
...
+  )
+  const current_level = next_level - 1
+  const current_level_experience = get_level_experience(current_level)
+
+  client.write('experience', {
+    totalExperience: 0,
+    level: current_level,
+    experienceBar:
+      (player_exp - current_level_experience) /
+      (next_level_experience - current_level_experience),
```
DeltaEvo, line 231: I think we should create a function for this like `current_level_experience`

src/player/experience.js
```diff
@@ -0,0 +1,233 @@
...
+  [92, 39803000],
+  [93, 41062000],
+  [94, 42341000],
+  [95, 43640000],
+  [96, 44959000],
+  [97, 46298000],
+  [98, 47657000],
+  [99, 49036000],
+  [100, 50435000],
+  [101, Infinity],
```
DeltaEvo, line 104: Do we need the 101

src/player/experience.js
```diff
@@ -0,0 +1,233 @@
...
+ * get the next level from the experience
+ * @param {number} exp
+ * @returns {[number, number]}
+ */
+function get_next_level_from_experience(exp) {
+  for (const [level, level_exp] of levels.entries()) {
+    if (level_exp > exp) {
+      return [level, level_exp]
+    }
+  }
```
DeltaEvo, line 172: Is this not just a `find` ?

src/player/experience.js
```diff
@@ -0,0 +1,233 @@
...
+  const { client } = state
+  const player_exp = get_experience(state)
+  const [next_level, next_level_experience] = get_next_level_from_experience(
+    player_exp
+  )
+  const current_level = next_level - 1
+  const current_level_experience = get_level_experience(current_level)
+
+  client.write('experience', {
+    totalExperience: 0,
```
DeltaEvo, line 227: Do you know for what this is used ? Because I don't see why the client need this

DeltaEvo (changes requested):

src/index.js
DeltaEvo: Here is fine

DeltaEvo:
Having jsdoc is great ! Thanks for adding it

src/index.js
```diff
@@ -46,12 +47,14 @@ function handle_login(world) {
       }),
       events: new EventEmitter(),
       gameMode: 2,
+      exp: 0,
```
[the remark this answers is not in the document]
Sceat: lint config would not let that pass :p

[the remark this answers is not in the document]
Vahelnir: because i apparently forgot to rename this. i'll correct that.

src/player/experience.js
```diff
@@ -0,0 +1,233 @@
+// TODO: choose between hardcoded/configured levels or generated as it was done before (generate_levels())
+const levels = new Map([
+  [0, 0],
```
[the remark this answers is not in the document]
Vahelnir: Probably not, tbh i copied this part blindly from the the java class

src/player/experience.js
```diff
@@ -0,0 +1,233 @@
...
+      911900,
+      1001000,
+      1100000,
+    ].map((value, index) => [index, value])
+  )
+  for (let i = 36; i < max_level + 1; i++) {
+    const previous_level = levels.get(i - 1)
+    const last = previous_level - levels.get(i - 2)
+    levels.set(i, previous_level + last + 20000)
+  }
```
Sceat, line 157: Prefer usage of mappers and avoid for loops

src/player/experience.js
Vahelnir: yeah but since level.entries() is returning an iterator, i thought that was more efficient that using an `Array.from()`. 
I'll remove the map part of the code so that will transform into a find anyway.

src/player/experience.js
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/**
+ * get the level and its min experience
+ * @param {number} level
+ * @returns {number}
+ */
+function get_level_experience(level) {
+  return levels.get(level)
+}
```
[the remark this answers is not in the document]
Sceat: that's very java friendly :ris:

src/player/experience.js
Vahelnir: Nope, i don't know what totalExperience is used for.
https://wiki.vg/Protocol#Set_Experience only says that we should check at the part of the wiki that explains the minecraft levels.

Sceat (changes requested):
:sparkles: sparkles rains, good FP effort tho, forget about getters and setters, take an input, produce a new output and that's it :grin: 

Experiences levels should imo be made through an algorithm instead of raw values like we add, with some hardcoded increasing layers

src/player/experience.js
Sceat: > more efficient
We don't do efficiency in js :grimacing: just porn code

src/player/experience.js
Vahelnir: That code will be removed anyway, but what do you mean by mappers ?

Sceat: the goal here seems to insert values into something, meaning you transform `[x]` into `[y]`
instead of iterating with a for loop you could `.map` x into y

```js
const foo = []
for (let i = 0; i < 10; i++) {
	foo.push(i)
}
```
vs
```js
const increment = (value, index) => index + 1
const foo = Array(10).fill(1).map(increment)
```

Vahelnir: Alright, that's what i thought.
Thanks !

src/player/experience.js
DeltaEvo: I will check the client sources then

src/player/experience.js
Vahelnir: Not sure but i think that it just make it easier because if your experience is higher than `50435000` for example, then the find will return -1. But i could return `[levels.length,Infinity]`if the find returns -1.

DeltaEvo: find will return undefined, and I think it's better to handle this when you do a find so if your find return undefined you handle the case and not have a non existing level 101

src/experience.js
```diff
@@ -0,0 +1,163 @@
...
+  const current_level = next_level - 1
+  const current_level_experience = get_level_experience(current_level)
+  return {
+    current_level,
+    current_level_progress:
+      (experience - current_level_experience) /
+      (next_level_experience - current_level_experience),
+    next_level_experience,
+  }
+}
```
Vahelnir, line 146: > I think we should create a function for this like `current_level_experience`
Is that better or do you prefer to extract the `level_progress` calculation outside of that function ? @DeltaEvo

DeltaEvo: I prefer it to be extracted in it's own function

src/player/experience.js
DeltaEvo: So it's doesn't seems to be used in the 1.16 client

src/experience.js
```diff
@@ -0,0 +1,199 @@
...
+ * @param {number} experience
+ * @returns {{level: number, remaining_experience: number}}
+ */
+export function experience_to_level(total_experience) {
+  const current_level = get_current_level(total_experience)
+  const current_level_experience = get_level_experience(current_level)
+  return {
+    level: current_level,
+    remaining_experience: total_experience - current_level_experience,
+  }
```
DeltaEvo, line 164: inline `get_current_level` and `get_level_experience`

src/experience.js
```diff
@@ -0,0 +1,199 @@
...
+  const current_level_experience = get_level_experience(level)
+  if (level + 1 >= levels.length) {
+    return 0
+  }
+  const next_level_experience = get_level_experience(level + 1)
+  return get_level_experience_progress(
+    remaining_experience,
+    current_level_experience,
+    next_level_experience
+  )
```
DeltaEvo, line 182: Inline `get_level_experience` and `get_level_experience_progress`

src/experience.js
```diff
@@ -0,0 +1,199 @@
...
+ * @param {any} state
+ */
+export function update_experience(state) {
+  const { client, experience: total_experience } = state
+  const { level, experience } = experience_to_level(total_experience)
+  const progress = level_progress({ level, experience })
+
+  client.write('experience', {
+    totalExperience: total_experience,
+    level: level,
```
DeltaEvo, line 196: Use property shorthand

src/index.js
```diff
@@ -10,6 +10,7 @@ import { register_villagers, spawn_villager } from './trade/spawn_villager.js'
 import { register_trades, open_trade } from './trade/trade.js'
 import { register_mobs, spawn_mob } from './mobs/spawn_mob.js'
 import { send_resource_pack } from './resource_pack.js'
+import { update_experience as player_update_experience } from './experience.js'
```
DeltaEvo, line 13: I don't think we need to rename here

DeltaEvo (changes requested):

DeltaEvo (approved):

Sceat (approved):

PR #38, opened 2020-11-05
Send a Boss Bar to the Player

Sauronox:


src/boss_bar.js
```diff
@@ -0,0 +1,77 @@
+import UUID from 'uuid-1345'
+
+export const colorBossBar = {
```
Sceat, line 3: Prefer using snakecase on function and variables names

src/boss_bar.js
```diff
@@ -0,0 +1,77 @@
...
+ *   dividers: typeDivisionBossBar.NO_DIVISION,
+ *   flags: flagsTypeBossBar.CREATE_FOG
+ * }
+ * ```
+ * 
+ * @param {object} state - The state must contain `{client}`.
+ * @param {object} params - The params may contain `{action, title, health, color, dividers, flags}`.
+ */
+export function display_boss_bar_to_player(state, params) {
+    const { client } = state
```
Sceat, line 60: When you only take one property prefer destructuring directly in the parameters, with a fallback
```js
({ client } = {}, params)
```

src/boss_bar.js
```diff
@@ -0,0 +1,77 @@
...
+    let { action, title, health, color, dividers, flags } = params
```
Sceat, line 61: this is a `const`, we *almost* never use `let`

src/boss_bar.js
```diff
@@ -0,0 +1,77 @@
...
+
+    if (action)
```
Sceat, line 63:
```suggestion
    if (!action)
```

src/boss_bar.js
```diff
@@ -0,0 +1,77 @@
...
+        return console.error('[Boss Bar] Action is not defined')
+
+    const configBossBar = {
+        entityUUID: UUID.v4(),
+        action,
+        title: JSON.stringify(title),
```
Sceat, line 69: you don't check for `title` presence, are you sure the packet accept `undefined` as a value ? you may want to fallback to an empty string

src/boss_bar.js
```diff
@@ -0,0 +1,77 @@
...
+ * @param {object} params - The params may contain `{action, title, health, color, dividers, flags}`.
+ */
+export function display_boss_bar_to_player(state, params) {
+    const { client } = state
+    let { action, title, health, color, dividers, flags } = params
+
+    if (action)
+        return console.error('[Boss Bar] Action is not defined')
+
+    const configBossBar = {
```
Sceat, line 66: we usually call that `options`

Sceat (changes requested):
Looking good :sparkles:, having a lot of different options you should use a unique provider object

```js
export default {
	Colors   : { PINK: 0, ... },
	Divisions: { ... },
	Types    : { ... },
	Actions  : { ... },
	display  : ({ state, options }) => {
		...
	}
}
```

i would even merge available options inside the parameter like

```js
display = ({ state, action, title, ... })
```

```js
import Bossbar from ''

Bossbar.display({
	action: Bossbar.Actions.ADD
})
```

src/boss_bar.js
```diff
@@ -0,0 +1,74 @@
...
+     *   flags: Bossbar.Flags.CREATE_FOG
+     * }
+     * ```
+     * @param {State} state
+     * @param {Option} opts
+     */
+    display({ state, options }) {
+        const { action, title, health, color, dividers, flags } = options
+
+        if (!action && action != 0)
```
Sceat, line 61: `if (action === undefined)`

src/boss_bar.js
```diff
@@ -0,0 +1,74 @@
...
+            return console.error('[Boss Bar] action is not defined')
+
+        state.client.write("boss_bar", {
+            entityUUID: uuid.v4(),
+            action,
+            title: JSON.stringify(title || { text: "Default Title" }),
+            health: health || 1.0,
+            color: color || 6,
+            dividers: dividers || 0,
+            flags: flags || 2
```
Sceat, line 71: prefer using the nullish coascelling operator instead of `||` which is now a deprecated practice, in case you have a falsy value it will trigger your fallback and this may be unwanted

src/boss_bar.js
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @typedef Title
+ * @property {string} text - Required
+ * @property {boolean} bold - Optional 
+ * @property {boolean} italic - Optional 
+ * @property {boolean} underlined - Optional 
+ * @property {boolean} strikethrough - Optional 
+ * @property {boolean} obfuscated - Optional
+ * @property {string} color - Optional 
+ * @property {array<object>} extra - Optional
```
Sceat, line 27: optionals in jsdoc are written `{string} [color]`

Sceat (changes requested):

src/boss_bar.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+export default {
+    Actions: { ADD: 0, REMOVE: 1, UPDATE_HEAlTH: 2, UPDATE_TITLE: 3, UPDATE_STYLE: 4, UPDATE_FLAGS: 5 },
+    Colors: { PINK: 0, BLUE: 1, RED: 2, GREEN: 3, YELLOW: 4, PURPLE: 5, WHITE: 6 },
+    Divisions: { NO_DIVISION: 0, '6_NOTCHES': 1, '10_NOTCHES': 2, '12_NOTCHES': 3, '20_NOTCHES': 4 },
+    Flags: { DRAKEN_SKY: 1, BOSS_BAR: 2, CREATE_FOG: 4 },
```
DeltaEvo, line 34: I know that @Sceat asked for this but can you change it to multiple names export, I prefer to let user choose if it want an object with all of this or do individual export ?
```js
export const Colors = { PINK: 0, ... }
export const Divisions = { ... }
export const Types = { ... }
export const Actions = { ... }
```

src/boss_bar.js
```diff
@@ -0,0 +1,74 @@
...
+     *   health: 0,
+     *   color: Bossbar.Colors.PINK,
+     *   dividers: Bossbar.Division.NO_DIVISION,
+     *   flags: Bossbar.Flags.CREATE_FOG
+     * }
+     * ```
+     * @param {State} state
+     * @param {Option} opts
+     */
+    display({ state, options }) {
```
DeltaEvo, line 58: `display` seems to be too generic for me I whould prefer something like `write_bossbar`

src/boss_bar.js
DeltaEvo, line 61: Why are you checking the action ? node-minecraft-protocol will do that himself

src/boss_bar.js
DeltaEvo, line 71: What are theses default values ?
I don't think we need to have default values

src/boss_bar.js
DeltaEvo, line 58: `{ state, options }` make this into two arguments `state` and `options` and destructure them

DeltaEvo (changes requested):

src/boss_bar.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+/**
+ * @typedef Option
+ * @property {number} action - ADD: 0, REMOVE: 1, UPDATE_HEAlTH: 2, UPDATE_TITLE: 3, UPDATE_STYLE: 4, UPDATE_FLAGS: 5 
+ * @property {Title} title - {text:"Exemple Title"}
+ * @property {float} health - min: 0.0 , max: 1.0 
+ * @property {number} color - PINK: 0, BLUE: 1, RED: 2, GREEN: 3, YELLOW: 4, PURPLE: 5, WHITE: 6
+ * @property {number} dividers -  NO_DIVISION: 0, '6_NOTCHES': 1, '10_NOTCHES': 2, '12_NOTCHES': 3, '20_NOTCHES': 4
+ * @property {number} flags - DRAKEN_SKY: 1, BOSS_BAR: 2, CREATE_FOG: 4
```
DeltaEvo, line 15: `DRAKEN_SKY` => `DARKEN_SKY`

src/boss_bar.js
Sceat: why that, single arguments is far more functional

DeltaEvo: Can you explain why ?

Sceat: Explain first

DeltaEvo: Because I prefer to have the state as first argument and an action as second one, to be coherant with the rest of the codebase and to have the same style as our reducers that will have the signature `(state, action): next_sate` like redux reducers

src/boss_bar.js
```diff
@@ -0,0 +1,72 @@
...
+ * ```
+ * @param {Parameter} 
+ */
+export const write_bossbar = ({ state, options }) => {
+    const { action, title, health, color, dividers, flags } = options
+
+    state.client.write("boss_bar", {
+        entityUUID: uuid.v4(),
+        action,
+        title: JSON.stringify(title),
```
DeltaEvo, line 66:
```suggestion
        title: title && JSON.stringify(title),
```

src/boss_bar.js
```diff
@@ -0,0 +1,66 @@
...
+    const { action, title, health, color, dividers, flags } = options
+
+    state.client.write("boss_bar", {
+        entityUUID: uuid.v4(),
+        action,
+        title: title && JSON.stringify(title),
+        health: health,
+        color: color,
+        dividers: dividers,
+        flags: flags
```
DeltaEvo, line 64: Use property shorthands http://es6-features.org/#PropertyShorthand

DeltaEvo (approved):

PR #44, opened 2020-11-08
basic chat system + welcome message

Zekirax:


src/chat.js
```diff
@@ -0,0 +1,17 @@
+export function sendChatWelcomeMsg(server, newClient) {
+  const msg = 'Bienvenue ' + newClient.username + ' !'
```
DeltaEvo, line 2: And use template strings https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/index.js
```diff
@@ -59,6 +60,12 @@ function handle_login(world) {
       throw error
     })
 
+    client.on('chat', function (packet) {
+      let msg = packet.message
+      msg = '<' + client.username + '> ' + msg
```
DeltaEvo, line 65: Use the chat format https://wiki.vg/Chat
And use template strings https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

DeltaEvo: You also have a translation component for this `"chat.type.text": "<%s> %s"`

DeltaEvo (changes requested):

src/chat.js
```diff
@@ -0,0 +1,45 @@
+export function send_chat_msg(server, json_msg, sender) {
```
DeltaEvo, line 1: Make a function with that signature `send_chat_msg({ server }, { message, sender })`

src/index.js
```diff
@@ -72,6 +73,7 @@ function handle_login(world) {
     spawn_villager(state)
     open_trade(state)
     spawn_mob(state)
+    add_chat_listener(server, client)
```
DeltaEvo, line 76: Rename this to `chat` and only take `state` as argument (add the server to the state if needed)

src/chat.js
DeltaEvo: And name this `write_chat_msg`

DeltaEvo (changes requested):
Almost there :)
Remove the `master` file that you added and remove all theses newlines

src/chat.js
Sceat: you should indeed avoid using many arguments and prefer an object instead to prevent parameters ordering ambiguities, 2 parameters is somehow a limit

src/chat.js
```diff
@@ -0,0 +1,30 @@
...
+      message: json_msg,
+      position: 0,
+      sender: sender.uuid,
+    })
+  }
+}
+
+export function chat(state) {
+  const server = state.server
+  const client = state.client
```
DeltaEvo, line 13: Destructure `state`

src/chat.js
```diff
@@ -0,0 +1,35 @@
+export function write_chat_msg(server, chat_message) {
```
DeltaEvo, line 1: Destructure `chat_message`

DeltaEvo: And take `state` instead of server the function should look like this `write_chat_msg({ server }, { message, sender })`

Zekirax: But why {server} and not just server, and why {message, sender} and not message, sender ?

DeltaEvo: For the first one is just to keep the codebase style and be more modular for the future, the server come from the state so instead of just taking the server take the whole state and take the server from that so in the future if we make a refactor and rename server or need to use something else for example the current chunk we only need to refactor the function and not all the caller and the whole codebase do this

For the second argument it's to create named arguments in javascript so the order don't matter `write_chat_message(a, b, c)` is not the same as `write_chat_message(a, c, b)` but `write_chat_message(a, { b, c })` is the same as `write_chat_message(a, { c, b })` and about the name `sender` and `message` for the arguments it's to match the packet names

src/chat.js
```diff
@@ -0,0 +1,31 @@
+export function write_chat_msg(server, { json_msg, client }) {
+  for (const other_client of Object.values(server.clients)) {
+    other_client.write('chat', {
+      message: json_msg,
+      position: 0,
+      sender: client.uuid,
+    })
+  }
+}
```
Sceat, line 9: A code style i'd like to see for these kind of behavior is to write a mapper and then apply it. That way when we read the code we don't have to pay attention at the loop, the brain can focus on the behavior alone

```suggestion
export function write_chat_msg(server, { json_msg, client: { uuid } }) {
  const options = {
    message: json_msg,
    position: 0, // use an enum here, 0 is a magic value
    sender: uuid,
  }
  const send_packet = client => client.write('chat', options)
  Object.values(server.clients).forEach(send_packet)
}
```

src/chat.js
```diff
@@ -0,0 +1,31 @@
...
+
+export function chat(state) {
+  const { server, client } = state
+
+  client.on('chat', function (packet) {
```
Sceat, line 14:
```suggestion
  client.on('chat', packet => {
```

src/chat.js
```diff
@@ -0,0 +1,31 @@
...
+  for (const other_client of Object.values(server.clients)) {
+    other_client.write('chat', {
+      message: json_msg,
+      position: 0,
+      sender: client.uuid,
+    })
+  }
+}
+
+export function chat(state) {
```
Sceat, line 11: this function name is a bit too simple, what does it do ? prefer using self-explaining names like `handle_chat` `initialize_chat` etc

Sceat (changes requested):

src/chat.js
DeltaEvo: Can you rename json_msg into message ?

src/chat.js
DeltaEvo: chat is fine for me

src/chat.js
```diff
@@ -0,0 +1,31 @@
...
```
DeltaEvo, line 11: Destructure `state` inside the argument

src/chat.js
```diff
@@ -0,0 +1,31 @@
+export function write_chat_msg({ server, json_msg, client }) {
```
DeltaEvo, line 1: Make the function have this signature `write_chat_msg({ server }, { message, sender })`

DeltaEvo (changes requested):

DeltaEvo:
And also remove the empty `master` file that you pushed

src/chat.js
```diff
@@ -0,0 +1,38 @@
+export const Position = {
+  CHAT: 0, // appears in the chat box
+  SYSTEM_MESSAGE: 1, // appears in the chat box
+  GAME_INFO: 2, // appears above the hotbar
+}
+
+export function write_chat_msg(
+  { server: { clients } },
+  { json_msg, client: { uuid } }
```
DeltaEvo, line 9: Name json_msg message

src/chat.js
```diff
@@ -0,0 +1,38 @@
...
+  const options = {
+    message: json_msg,
+    position: Position.CHAT,
+    sender: uuid,
+  }
+  const send_packet = (client) => client.write('chat', options)
+  Object.values(clients).forEach(send_packet)
+}
+
+export function chat({ state: { server, client } }) {
```
DeltaEvo, line 20: Take state as parameter directly instead of an object with state so { server, client } directly

DeltaEvo (changes requested):
A last few things and you still have a empty `master`file in your branch, can your remove it please ?

DeltaEvo (approved):

PR #53, opened 2020-11-15
Chat show item

Zekirax:
Can show an item in chat

### DIFF
--- a/src/chat.js
+++ b/src/chat.js
@@ -1,7 +1,39 @@
+import nbt from 'prismarine-nbt'
+import minecraftData from 'minecraft-data'
+
+import { item_to_slot } from './items.js'
+import { version } from './settings.js'
+import logger from './logger.js'
+
+const mcData = minecraftData(version)
+const log = logger(import.meta)
+
 export const Position = {
   CHAT: 0, // appears in the chat box
   SYSTEM_MESSAGE: 1, // appears in the chat box
   GAME_INFO: 2, // appears above the hotbar
+}
+
+function is_command_function(message) {
+  return message.trimStart()[0] === '/'
+}
+
+export function slot_to_chat({ nbtData, itemCount, itemId }) {
+  const tag = nbt.simplify(nbtData)
+  const { name } = mcData.items[itemId]
+
+  const chat = {
+    ...JSON.parse(tag.display.Name),
+    hoverEvent: {
+      action: 'show_item',
+      contents: {
+        id: name,
+        itemCount,
+        tag: JSON.stringify(tag),
+      },
+    },
+  }
+  return chat
 }
 
 export function write_chat_msg(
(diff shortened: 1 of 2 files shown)


src/chat.js
```diff
@@ -17,20 +29,57 @@ export function write_chat_msg(
...
+            ...JSON.parse(tag.display.Name),
+            hoverEvent: {
+              action: 'show_item',
+              contents: {
+                id: name,
+                itemCount,
+                tag: JSON.stringify(tag),
+              },
+            },
+          }
```
DeltaEvo, line 64: Create a function `slot_to_chat` for this

src/chat.js
```diff
@@ -1,9 +1,21 @@
...
+const log = logger(import.meta)
+
 export const Position = {
   CHAT: 0, // appears in the chat box
   SYSTEM_MESSAGE: 1, // appears in the chat box
   GAME_INFO: 2, // appears above the hotbar
 }
 
+function is_command_function(message) {
+  if (message.trimStart()[0] === '/') return true
```
DeltaEvo, line 17:
```suggestion
  return message.trimStart()[0] === '/'
```

src/chat.js
```diff
@@ -17,20 +29,57 @@ export function write_chat_msg(
...
   client.on('chat', (packet) => {
+    if (is_command_function(packet.message)) {
+      log.debug({ sender: client.uuid, command: packet.message }, 'Command')
+
+      return
+    } else {
+      log.debug({ sender: client.uuid, message: packet.message }, 'Message')
+    }
+    const splited_message = packet.message.split(/(%item\d%)/)
+    const formatted_message = splited_message.map((part) => {
```
DeltaEvo, line 42: I think we can remove splitted_message and have directly
```suggestion
    const formatted_message = packet.message.split(/(%item\d%)/).map((part) => {
```

src/chat.js
```diff
@@ -17,20 +29,57 @@ export function write_chat_msg(
   Object.values(clients).forEach(send_packet)
 }
 
-export default function chat({ server, client }) {
+export default function chat({ server, client, inventory, world }) {
   client.on('chat', (packet) => {
+    if (is_command_function(packet.message)) {
```
DeltaEvo, line 34: Since we are using `message` a lot maybe we can destructurate `packet` ?

src/chat.js
```diff
@@ -17,20 +29,57 @@ export function write_chat_msg(
...
+                id: name,
+                itemCount,
+                tag: JSON.stringify(tag),
+              },
+            },
+          }
+          return chat
+        } else {
+          return { text: part }
+        }
```
DeltaEvo, line 68: You can remove this `else` and have a `return { text: part }` at the end of the function

src/chat.js
```diff
@@ -17,20 +29,57 @@ export function write_chat_msg(
...
+      } else {
+        return { text: part }
+      }
```
DeltaEvo, line 71: Same as comment above ^

DeltaEvo (changes requested):

src/chat.js
```diff
@@ -1,9 +1,41 @@
...
   SYSTEM_MESSAGE: 1, // appears in the chat box
   GAME_INFO: 2, // appears above the hotbar
 }
 
+function is_command_function(message) {
+  return message.trimStart()[0] === '/'
+}
+
+export function slot_to_chat({ count, type }, { items }) {
+  const { nbtData, itemCount, itemId } = item_to_slot(items[type], count)
```
DeltaEvo, line 21: If you are calling item_to_slot then this function is a `item_to_chat`, this function should take { nbtData, itemCount, itemId } as parameter and you should remove the `item_to_slot` call

DeltaEvo (approved):

PR #55, opened 2020-11-18
Add basic scoreboard system with animations

Upd4ting:


src/index.js
```diff
@@ -18,6 +18,7 @@ import {
   reduce_plugin_channels,
   transform_plugin_channels,
 } from './plugin_channels.js'
+import { init_scoreboard } from './scoreboard/init_scoreboard.js'
```
DeltaEvo, line 21: `init_` seems to not add any meaning, `scoreboard` seems like a good name

src/scoreboard/animations.js
```diff
@@ -0,0 +1,346 @@
...
+
+function isFormatCode(color) {
+  return (
+    color === '§k' ||
+    color === '§l' ||
+    color === '§m' ||
+    color === '§n' ||
+    color === '§o'
+  )
+}
```
DeltaEvo, line 21: We should not use theses magic color values, instead use the `color` property inside the https://wiki.vg/Chat Component, it's more readable, and it's future proof for example you can have `rgb` colors inside this property

src/scoreboard/init_scoreboard.js
```diff
@@ -0,0 +1,111 @@
...
+    add_sideline({ client }, { text: 'Crit §7: §4§oBientôt' })
+    add_sideline({ client }, { text: 'Chance §7: §4§oBientôt' })
+    add_sideline({ client }, { text: '' })
+    add_sideline({ client }, { text: 'Métiers §4§oBientôt' })
+    add_sideline({ client }, { text: '§7- §aAucun' })
+    add_sideline({ client }, { text: '' })
+    add_sideline({ client }, { text: 'Or §7: §4§oBientôt' })
+    add_sideline({ client }, { text: '' })
+    add_sideline({ client }, { text: 'www.aresrpg.fr' })
+    flush_sideline({ client })
```
DeltaEvo, line 36: I think this should be done with online one `send_sideline` function

src/scoreboard/init_scoreboard.js
```diff
@@ -0,0 +1,111 @@
...
+
+export function init_scoreboard({ client, events }) {
+  events.once('state', () => {
+    init_sidebar({ client })
+    setname_sidebar(
+      { client },
+      {
+        name: '§6Statistiques',
+      }
+    )
```
DeltaEvo, line 18: I think this should be done with only one `send_sidebar` function

src/scoreboard/scoreboard.js
```diff
@@ -0,0 +1,74 @@
...
+) {
+  const obj = {
+    name,
+    action,
+    displayText: JSON.stringify(displayText),
+    type,
+  }
+
+  client.write('scoreboard_objective', obj)
+}
```
DeltaEvo, line 41: Do you really need a function for that ?

src/scoreboard/scoreboard.js
```diff
@@ -0,0 +1,74 @@
...
+) {
+  const obj = {
+    scoreName,
+    action,
+    itemName,
+    value,
+  }
+
+  client.write('scoreboard_score', obj)
+}
```
DeltaEvo, line 60: Do you really need a function for that

src/scoreboard/scoreboard.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {{position: number, name: string}} Options
+ */
+export function scoreboard_display_objective({ client }, { position, name }) {
+  const obj = {
+    position,
+    name,
+  }
+
+  client.write('scoreboard_display_objective', obj)
+}
```
DeltaEvo, line 74: Do you really need a function for that ?

src/scoreboard/sidebar.js
```diff
@@ -0,0 +1,172 @@
...
+  scoreboard_objective,
+  scoreboard_score,
+  scoreboard_display_objective,
+} from './scoreboard.js'
+
+import { update_basetext_animation } from './animations.js'
+
+//TODO: should be moved to redis. not stateless if we clusterize the server.
+const clientLinesMap = new Map()
+const clientTitleMap = new Map()
```
DeltaEvo, line 15: `TODO: should not exist :upside_down_face:`, you should not have global state like that
I think your issue here is that your breaking the sidebar into multiple function calls
Here for me you should have only one function
`update_sidebar` that take the old sidebar and the new sidebar and does what is needed to create/update/delete it

src/scoreboard/sideline.js
```diff
@@ -0,0 +1,118 @@
...
+  init_sidebar,
+  send_sidebar,
+  clear_sidebar,
+  setline_sidebar,
+  removeline_sidebar,
+} from './sidebar.js'
+
+//TODO: should be moved to redis. not stateless if we clusterize the server.
+const clientOldMap = new Map()
+const clientBufferMap = new Map()
```
DeltaEvo, line 11: Same as `sidebar`

src/scoreboard/animations.js
```diff
@@ -0,0 +1,346 @@
+import { set_sideline, get_sideline } from './sideline.js'
+import { setname_sidebar, getname_sidebar } from './sidebar.js'
+
+//TODO: should be moved to redis. not stateless if we clusterize the server.
+const clientAnimationMap = new Map()
```
DeltaEvo, line 5: This should not be needed, you should not break the logic into multiple function and use a global state, just make a function that create + start the animation and that return a function to cancel it like
```js
function start_animation(...) {
	const id = setInterval(...)
    return {
        cancel() { clearInterval(id) }
    }
}
```

src/scoreboard/animations.js
```diff
@@ -0,0 +1,346 @@
...
+    setname_sidebar({ client }, { name: text, animated: true })
+  } else {
+    set_sideline({ client }, { line: animation.line, text, animated: true })
+  }
+}
+
+function process_animation({ client }, { animation }) {
+  const effect = Array.isArray(animation.effect)
+    ? animation.effect
+    : [animation.effect]
```
DeltaEvo, line 34: Simplify this make animation.effet to always be an array

DeltaEvo (changes requested):

DeltaEvo (changes requested):
There is too much functions too call to use this, I think we should have only one function to create and send a sidebar scoreboard/sideline scoreboard/an animation
The animation code seems nice to use but the implementation is too complex for me, but first we should simplify the interface before the implementation

src/scoreboard/animations.js
Upd4ting: At the beginning I wanted to use the chat component but then I saw that the scoreboard title handle chat component but not scoreboard line. As I wanted to make the animation for both line and title, I used these magic color values.

I can handle the different cases and let us use chat component only for the title, but it makes the implementation of animation more complex.

DeltaEvo: I looked at minecraft code and indead it dosen't parse a component for this but it parse a chat component for the team prefix and suffix
We can hack something with the teams but I don't like this idea
But I think we can just make a function that transform a chat component into a string and throw an Error if it's not possible

Upd4ting: So the scoreboard interface will take a chat component and will convert that to a string for internal use?

DeltaEvo: Yes, but for the magic values you will add them into a list or an enum so they will not be magic values anymore :)

Upd4ting: Perfect 👍

src/scoreboard/sidebar.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    const displayObjective = {
+      position: Positions.SIDEBAR,
+      name: client.username,
+    }
+
+    client.write('scoreboard_objective', objective)
+    client.write('scoreboard_display_objective', displayObjective)
+  }
```
DeltaEvo, line 76:
```suggestion
  if (last.title !== next.title) {
    const name = client.username
    
  	if (next.title === undefined) {
  	  client.write('scoreboard_objective', { name, action: ObjectiveActions.REMOVE })
  	} else {
  	  const displayText = JSON.stringify(next.title)
  	  const type = Types.INTEGER
  	  
  	  if (last.title === undefined) {
  	      client.write('scoreboard_objective', { name, displayText, type, action: ObjectiveActions.CREATE })
  	      client.write('scoreboard_display_objective', {
            position: Positions.SIDEBAR,
            name,
          })
  	  } else
  	    client.write('scoreboard_objective', { name, displayText, type, action: ObjectiveActions.UPDATE })
  	}
  }
```

src/scoreboard/sidebar.js
```diff
@@ -0,0 +1,114 @@
...
+
+  const itemsToDelete = lastLines.filter((item) => {
+    const sameLine = nextLines.find((item2) => item.line == item2.line)
+    return !sameLine || sameLine.text != item.text
+  })
+  const itemsToUpdate = nextLines.filter(
+    (item) =>
+      !lastLines.find(
```
DeltaEvo, line 84: Use `some` instead of find https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/some

src/scoreboard/sidebar.js
```diff
@@ -0,0 +1,114 @@
...
+      name: client.username,
+    }
+
+    client.write('scoreboard_objective', objective)
+    client.write('scoreboard_display_objective', displayObjective)
+  }
+
+  const itemsToDelete = lastLines.filter((item) => {
+    const sameLine = nextLines.find((item2) => item.line == item2.line)
+    return !sameLine || sameLine.text != item.text
```
DeltaEvo, line 80:
```suggestion
    return sameLine?.text !== item.text ?? true
```

src/scoreboard/util.js
```diff
@@ -0,0 +1,35 @@
...
+    if (chat.bold) text += ChatColor.BOLD
+    if (chat.strikethrough) text += ChatColor.STRIKETHROUGH
+    if (chat.underline) text += ChatColor.UNDERLINE
+    if (chat.italic) text += ChatColor.ITALIC
+
+    text += chat.text !== '' ? chat.text : ' '
+  }
+
+  return text
+}
```
DeltaEvo, line 35:
```suggestion
const to_array = value => Array.isArray(value) ? value : [value]

export function chat_to_text(chat) {
  return to_array(chat).map(part => {
  	const modifiers = [
  	  ChatColor[part.color?.toUpperCase()] ?? '',
  	  part.obfuscated ? ChatColor.OBFUSCATED : '',
  	  part.bold ? ChatColor.BOLD : '',
  	  part.strikethrough ? ChatColor.STRIKETHROUGH : '',
      part.underline ? ChatColor.UNDERLINE : '',
      part.italic ? ChatColor.ITALIC : ''
  	]
  	
  	const extra = part.extra ? chat_to_text(part.extra) : ''
  	
  	return `${modifiers.join('')}${part.text}${extra}${ChatColor.RESET}`
  }).join('')
}
```

src/scoreboard/enums.js
```diff
@@ -0,0 +1,53 @@
...
+  STRIKETHROUGH: '§m',
+  UNDERLINE: '§n',
+  ITALIC: '§o',
+  RESET: '§r',
+  isFormatCode: (color) =>
+    color === '§k' ||
+    color === '§l' ||
+    color === '§m' ||
+    color === '§n' ||
+    color === '§o',
```
DeltaEvo, line 52: I think we should move this outside of this object because it will be in the ways if we use Object.entries on this object to get all chat colors for example

DeltaEvo (changes requested):
Much better :+1: I will do a review of the animation system tomorrow

DeltaEvo:
For the animations code I think we can simplify it with this points:
-  I don't think we need a `start` function, you can just start the animation when we create it
-  I think it would be easier to just create all the frames an cycles through them for example
```js
```

src/scoreboard/animations.js
```diff
@@ -0,0 +1,194 @@
...
+            return [].concat(e)
+          }
+
+          const effect =
+            animation.effects[j - i - offset + animation.effects.length]
+          return []
+            .concat(effect || [])
+            .concat(e)
+            .concat(effect ? [ChatColor.RESET].concat(colors) : [])
+        })
```
DeltaEvo, line 183: I think this can be simplified if you keep the chat components, so you don't need to care about colors inside the text
What I'm thinking is implementing a `split_chat_component` function
```js
split_chat_component({ text: 'hello', color: 'red' }, '')
=> [{ text: 'h', color: 'red' }, { text: 'e', color: 'red' }, ...]
```
And an `optimize_chat_component` that regroup the split chat component

Sceat:
@Upd4ting from what i see on the reviews you may be interested in deep diving into [pure programming concepts](https://www.freecodecamp.org/news/functional-programming-principles-in-javascript-1b8fc6c3563f/) :+1:

Upd4ting:
@Sceat yeah thanks for the reading :)

Sceat (changes requested):
I propose that you move your animation system on top of the leaner scoreboard system (once https://github.com/aresrpg/aresrpg/pull/274 is merged)

PR #74, opened 2020-12-20
Interact with NPCs

ToufouMaster:
Right click on an NPC and talk with them!

Added a detect_dialog(server, { client, world }) Function
Added a speak_to(target, server, { client, world }) Function


src/index.js
```diff
@@ -20,6 +20,7 @@ import { reduce_position } from './position.js'
 import { online_mode, version } from './settings.js'
 import { register_traders, spawn_merchants } from './trade/spawn_villagers.js'
 import { open_trade, register_trades } from './trade/trade.js'
+import detect_dialog from './dialogs/dialog.js'
```
DeltaEvo, line 23: This should go in `mobs` folder I think

src/index.js
```diff
@@ -81,6 +82,7 @@ async function observe_client(context) {
   spawn_mob(context)
   spawn_merchants(context)
   open_trade(context)
+  detect_dialog(server, context)
```
DeltaEvo, line 85: Follow the same style as the other function by passing only one argument

src/dialogs/dialogs.js
```diff
@@ -0,0 +1,78 @@
...
+        '§fGwanno ereb nin !',
+        '§fNo dhínen !',
+        '§fGarich i dhôl goll o Orch',
+        '§fNai Aragog meditha le!',
+        '§fHeca, firimar !',
+      ],
+      linearity: false,
+    },
+  },
+}
```
DeltaEvo, line 78: Maybe this should go directly in `src/mobs/mobs.js` ?

src/dialogs/dialogs.js
```diff
@@ -0,0 +1,78 @@
...
+        "§fC'est toi " +
+          'PLAYER_NAME' +
+          ' ? je pensais que tu étais plus grand !',
+        "§fFais attention à ne pas t'aventurer dans les catacombes, c'est dangereux !",
+        "§fJe n'ai pas le temps de te parler, va t'en ou j'apelle la garde !",
+        '§fEncore toi ?',
+        '§fUn jour les dragons reviendront et ce sera la fin de notre monde !',
+        '§fLaisse moi !',
+      ],
+      linearity: false,
```
DeltaEvo, line 19: `linearity` here is always false, do we really want this functionality ?

src/dialogs/dialog.js
```diff
@@ -0,0 +1,68 @@
...
+        const x = Math.floor(Math.random() * mobdata[target].dialogs.length)
+        write_chat_msg(
+          { server },
+          {
+            message: JSON.stringify(
+              mobdata[target].name + ': ' + mobdata[target].dialogs[x]
+            ),
+            client,
+          }
+        )
```
DeltaEvo, line 64: This will send a message to all the players, I don't think you want that instead use `client.write('chat', ...)`

src/dialogs/dialog.js
```diff
@@ -0,0 +1,68 @@
...
+        dialog_text_id++
+      } else {
+        const x = Math.floor(Math.random() * mobdata[target].dialogs.length)
+        write_chat_msg(
+          { server },
+          {
+            message: JSON.stringify(
+              mobdata[target].name + ': ' + mobdata[target].dialogs[x]
+            ),
```
DeltaEvo, line 61: Please use the Object or Array format of a Chat Message https://wiki.vg/Chat

src/dialogs/dialog.js
```diff
@@ -0,0 +1,68 @@
...
+            message: JSON.stringify(
+              '[' +
+                (dialog_text_id + 1) +
+                '/' +
+                mobdata[target].dialogs.length +
+                '] ' +
+                mobdata[target].name +
+                ': ' +
+                mobdata[target].dialogs[dialog_text_id]
+            ),
```
DeltaEvo, line 49: Please use the Object or Array format of a Chat Message https://wiki.vg/Chat

src/dialogs/dialog.js
```diff
@@ -0,0 +1,68 @@
...
+  client.on('use_entity', ({ target, mouse, sneaking }) => {
+    if (mouse === right_click && sneaking === false) {
+      speak_to(target, server, { client, world })
+    }
+  })
+}
+
+export function speak_to(target, server, { client, world }) {
+  // TODO:Remove Server argument in Chat.js
+  const mob = floor1.mobs[target] // TODO: Replace with Client current World
```
DeltaEvo, line 20: This only works because `register_mobs` here is the first of the list https://github.com/aresrpg/aresrpg/blob/master/src/index.js#L32

src/dialogs/dialog.js
```diff
@@ -0,0 +1,68 @@
...
+  console.log(mob)
+  if (mob !== undefined) {
+    const mobdata = mobs[mob.mob]
```
DeltaEvo, line 23: I don't think `speak_to` should check that :thinking:, this should be done in `detect_dialog` I think

src/dialogs/dialog.js
```diff
@@ -0,0 +1,68 @@
+import { floor1 } from '../world.js'
+import { write_chat_msg } from '../chat.js'
+
+import { default_dialog_id, mobs } from './dialogs.js'
+
+let latest_target = default_dialog_id
+let dialog_text_id = 0
```
DeltaEvo, line 7: This should not be global, since this mean that it will break if we have multiple players

src/dialogs/dialog.js
```diff
@@ -0,0 +1,68 @@
...
+                (dialog_text_id + 1) +
+                '/' +
+                mobdata[target].dialogs.length +
+                '] ' +
+                mobdata[target].name +
+                ': ' +
+                mobdata[target].dialogs[dialog_text_id]
+            ),
+            client,
+          }
```
DeltaEvo, line 51: This will send a message to all the players, I don't think you want that instead use `client.write('chat', ...)`

src/dialogs/dialogs.js
```diff
@@ -0,0 +1,78 @@
+export const default_dialog_id = -1
+
+export const mobs = {
+  thebes_citizen: {
+    // Citoyen de Thebes
+    '-1': {
+      name: '§a Citoyen §7',
```
DeltaEvo, line 7: Can you not use the mob name for this ?

src/dialogs/dialogs.js
```diff
@@ -0,0 +1,78 @@
...
+      dialogs: [
+        "§fC'est incroyable ce beau temps !",
+        "§fC'est toi " +
+          'PLAYER_NAME' +
```
DeltaEvo, line 11: You should replace this PLAYER_NAME and I prefer `{player}` instead

src/dialogs/dialogs.js
```diff
@@ -0,0 +1,78 @@
+export const default_dialog_id = -1
+
+export const mobs = {
+  thebes_citizen: {
+    // Citoyen de Thebes
+    '-1': {
+      name: '§a Citoyen §7',
+      dialogs: [
+        "§fC'est incroyable ce beau temps !",
```
DeltaEvo, line 9: Can you try to avoid these magic color codes and instead use the color name in a chat component https://wiki.vg/Chat ?

DeltaEvo (changes requested):
Need some modifications

And i'm not convinced by the `linearity` feature I don't think we need that

src/dialogs/dialogs.js
```diff
@@ -0,0 +1,78 @@
+export const default_dialog_id = -1
+
+export const mobs = {
+  thebes_citizen: {
+    // Citoyen de Thebes
+    '-1': {
```
DeltaEvo, line 6: What is this `-1` ?

src/dialogs/dialogs.js
```diff
@@ -0,0 +1,78 @@
+export const default_dialog_id = -1
```
DeltaEvo, line 1: What is this ? I don't like this magic number

src/dialogs/dialog.js
ToufouMaster: Thanks i go back to code

src/dialogs/dialogs.js
ToufouMaster: if you want to create a character who told a little story to the player yep that's usefull
but if you don't it's useless

src/dialogs/dialogs.js
ToufouMaster: by what do you want it to be replaced?

src/dialogs/dialogs.js
ToufouMaster: -1 is global it allows to make the values ​​inside valid for all mobs
but if you add a "5917" and that you click on the mob with id 5917 the discussion launched will be the one named "5917"

src/dialogs/dialogs.js
ToufouMaster: it's the -1 in dialogs but you can change it

src/dialogs/dialogs.js
ToufouMaster: by what do you want it to be replaced?

DeltaEvo: I want that we use this https://github.com/aresrpg/aresrpg/blob/master/src/mobs/mobs.js#L428 instead

src/dialogs/dialogs.js
DeltaEvo: This `-1` is very weird, why not have choosen something like `default` ?
But I don't see this feature used, and id are not stable I don't like having id here
Since we don't use it I think this should be removed

src/index.js
ToufouMaster: Ok

src/dialogs/dialogs.js
ToufouMaster: removed

src/dialogs/dialog.js
ToufouMaster: Ok

src/dialogs/dialog.js
ToufouMaster: removed

src/dialogs/dialog.js
ToufouMaster: Ok

src/dialogs/dialogs.js
ToufouMaster: Ok

src/dialogs/dialogs.js
ToufouMaster: removed

src/dialogs/dialogs.js
ToufouMaster: removed

src/dialogs/dialogs.js
ToufouMaster: Ok

src/index.js
ToufouMaster: Ok

src/dialogs/dialog.js
ToufouMaster: I think it's Ok

src/dialogs/dialog.js
ToufouMaster: I Think it's Ok

src/dialogs/dialog.js
ToufouMaster: Ok

src/dialogs/dialogs.js
ToufouMaster: Ok

src/dialogs/dialog.js
ToufouMaster: an idea of how change that?

src/dialogs/dialogs.js
ToufouMaster: Ok
added in chat.js a const variable who has the 16 different symbol

src/mobs/dialog.js
```diff
@@ -0,0 +1,41 @@
...
+      translate: 'chat.type.text',
+      with: [
+        {
+          text: mobdata.name,
+        },
+        {
+          text: mobdata.dialogs[x],
+        },
+      ],
+    })
```
DeltaEvo, line 32: I don't know if we want to use `chat.type.text` here since it's for player chat
```suggestion
    const message = JSON.stringify([
      { text: mobdata.name, color: "green" },
      { text: ":", color: "gray" },
      { text: mobdata.dialogs[x], color: "white" },
    ])
```

src/mobs/dialog.js
```diff
@@ -0,0 +1,41 @@
...
+    const options = {
+      message,
+      position: Position.CHAT,
+      sender: client.uuid,
+    }
+    const send_packet = (client) => client.write('chat', options)
+    Object.values([client]).forEach(send_packet)
+  }
```
DeltaEvo, line 40: Object.values([client]).forEach(send_packet) is the same as `send_packet(client) so I think we should just write
```suggestion
    client.write('chat', {
      message,
      position: Position.CHAT,
      sender: client.uuid,
    })
  }
```

src/mobs/mobs.js
```diff
@@ -1,3 +1,78 @@
...
+      chat_color.white +
+        "C'est toi {player} ? Je pensais que tu étais plus grand !",
+      chat_color.white +
+        "Fais attention à ne pas t'aventurer dans les catacombes, c'est dangereux !",
+      chat_color.white +
+        "Je n'ai pas le temps de te parler, va t'en ou j'apelle la garde !",
+      chat_color.white + 'Encore toi ?',
+      chat_color.white +
+        'Un jour les dragons reviendront et ce sera la fin de notre monde !',
+      chat_color.white + 'Laisse moi !',
```
DeltaEvo, line 17: For the color this is not what I was thinking, you have two ways to have colors within chat, one with magic code (what you are using) and one with color names like in my comment above (`{ text: mobdata.name, color: "green" }`)

Here all the lines have the same color so I think we should remove the color from here and apply it on top, like in my suggestion :)

src/mobs/mobs.js
```diff
@@ -1,3 +1,78 @@
+import { chat_color } from '../chat.js'
+
+export const mobs_dialogs = {
```
DeltaEvo, line 3: By moving inside `mobs.js` I was more thinking about adding the dialogs into the current structure so for example change
```js
  thebes_citizen: {
    displayName: 'Citoyen de Thebes',
    type: 'npc',
    mob: 'villager',
  },
```
Into
```js
  thebes_citizen: {
    displayName: 'Citoyen de Thebes',
    type: 'npc',
    mob: 'villager',
    dialogs: [
      "C'est incroyable ce beau temps !",
      "C'est toi {player} ? Je pensais que tu étais plus grand !",
      "Fais attention à ne pas t'aventurer dans les catacombes, c'est dangereux !",
      "Je n'ai pas le temps de te parler, va t'en ou j'apelle la garde !",
      'Encore toi ?',
      'Un jour les dragons reviendront et ce sera la fin de notre monde !',
      'Laisse moi !'
    ]
  },
```

src/mobs/dialog.js
```diff
@@ -0,0 +1,41 @@
...
+  if (mobdata !== undefined) {
+    const x = Math.floor(Math.random() * mobdata.dialogs.length)
+    const message = JSON.stringify({
+      translate: 'chat.type.text',
+      with: [
+        {
+          text: mobdata.name,
+        },
+        {
+          text: mobdata.dialogs[x],
```
DeltaEvo, line 29: You need to replace `{player}` by the player name (client.username)

src/mobs/dialog.js
```diff
@@ -0,0 +1,41 @@
+import { floor1 } from '../world.js'
+import { Position } from '../chat.js'
+
+import { mobs, mobs_dialogs } from './mobs.js'
+
+export default function detect_dialog({ client }) {
```
DeltaEvo, line 6: The `detect` part of `detect_dialog` name bother me, I think just `dialog` is fine

DeltaEvo (changes requested):

src/mobs/dialog.js
ToufouMaster: how?

DeltaEvo: Using https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/String/replace

src/mobs/dialog.js
```diff
@@ -0,0 +1,41 @@
...
+  const right_click = 2
+  client.on('use_entity', ({ target, mouse, sneaking }) => {
+    if (mouse === right_click && sneaking === false) {
+      const mob = floor1.mobs[target]
+      if (mob !== undefined) {
```
DeltaEvo, line 11:
```suggestion
      if (target in floor1.mobs) {
        const mob = floor1.mobs[target]
```

src/mobs/dialog.js
ToufouMaster: Ok

src/mobs/dialog.js
ToufouMaster: ok

src/mobs/dialog.js
ToufouMaster: Ok

src/mobs/dialog.js
ToufouMaster: ok

src/mobs/dialog.js
ToufouMaster: ok

src/mobs/mobs.js
ToufouMaster: ok

src/mobs/mobs.js
ToufouMaster: ok

src/chat.js
```diff
@@ -4,6 +4,25 @@ export const Position = {
...
+  dark_gray: '§8',
+  blue: '§9',
+  bright_green: '§a',
+  cyan: '§b',
+  red: '§c',
+  pink: '§d',
+  yellow: '§e',
+  white: '§f',
+}
+
```
DeltaEvo, line 25: I think we can remove that since it's not used anymore

src/mobs/dialog.js
```diff
@@ -0,0 +1,35 @@
...
+        color: 'white',
+      },
+    ])
+    client.write('chat', {
+      message,
+      position: Position.CHAT,
+      sender: client.uuid,
+    })
+  }
+}
```
DeltaEvo, line 35: `mobs[mob.mob]` is repeated a lot I think you should destructure it at the top of the function like that
```js
const { dialogs, displayName } = mobs[mob.mob]
```

src/mobs/dialog.js
```diff
@@ -0,0 +1,35 @@
...
+import { Position } from '../chat.js'
+
+import { mobs } from './mobs.js'
+
+export default function dialog({ client }) {
+  const right_click = 2
+  client.on('use_entity', ({ target, mouse, sneaking }) => {
+    if (mouse === right_click && sneaking === false) {
+      if (target in floor1.mobs) {
+        const mob = floor1.mobs[target]
```
DeltaEvo, line 11: Can you add a FIXME ?
```suggestion
        const mob = floor1.mobs[target] // FIXME: check proper entity id here
```

DeltaEvo (changes requested):
Last round of comments and I think it will be good :+1:

src/chat.js
ToufouMaster: Yep
Ok

src/index.js
```diff
@@ -11,6 +11,7 @@ import { scan } from './iterables.js'
 import logger from './logger.js'
 import login from './login.js'
 import { register_mobs, spawn_mob } from './mobs/spawn_mob.js'
+// import { drop_item, collect_item } from './mobs/item_drop.js'
```
DeltaEvo, line 14: Remove this :)

src/index.js
```diff
@@ -74,13 +75,13 @@ function transform_action(action) {
 async function observe_client(context) {
   /* Observers that handle the protocol part.
    * They get the client and should map it to minecraft protocol */
-
```
DeltaEvo, line 77: Read this line

src/index.js
```diff
@@ -142,6 +144,8 @@ pipeline(
         },
       })
 
+      actions.write('update_time', { age: 0, time: -1 })
+
```
DeltaEvo, line 148: Remove that :)

src/index.js
```diff
@@ -141,7 +143,6 @@ pipeline(
           actions.write({ type, payload })
         },
       })
-
```
DeltaEvo, line 144: Re add this line

DeltaEvo (approved):

PR #79, opened 2020-12-24
Add equipment to mobs

loucass003:


src/mobs/mobs.js
```diff
@@ -3,6 +3,9 @@ export const mobs = {
     displayName: 'Abrakleur',
     type: 'mob',
     mob: 'enderman',
+    equipments: {
+      0: { type: 'menitrass_100', count: 1 },
```
DeltaEvo, line 7: Can you use more meaningfull names instead of 0 like `main_hand` ?

DeltaEvo (changes requested):

DeltaEvo (approved):

PR #93, opened 2020-12-28
World Border

ToufouMaster:
Added Red_screen Control Function

### DIFF
--- a/src/login.js
+++ b/src/login.js
@@ -56,7 +56,7 @@ export default function login({ client, events }) {
       chunkZ: chunk.z,
     })
 
-    set_world_border({ client, x: 512, z: 510, radius: 10000, speed: 1 })
+    set_world_border({ client, x: 512, z: 512, radius: 1020, speed: 1 })
 
     load_chunks(state, { client, events, chunks: [chunk] })
 
--- a/src/login.js
+++ b/src/login.js
@@ -56,7 +56,7 @@ export default function login({ client, events }) {
       chunkZ: chunk.z,
     })
 
-    set_world_border({ client, radius: 10000 })
+    set_world_border({ client, x: 512, z: 510, radius: 10000, speed: 1 })
 
     load_chunks(state, { client, events, chunks: [chunk] })
 
--- a/src/login.js
+++ b/src/login.js
@@ -4,6 +4,7 @@ import { write_brand } from './plugin_channels.js'
 import { dimension_codec, overworld } from './world/codec.js'
 import { load_chunks } from './chunk/update.js'
 import { write_title } from './title.js'
+import { set_world_border, red_screen } from './lowlife.js'
 
 export default function login({ client, events }) {
   events.once('state', (state) => {
@@ -55,6 +56,9 @@ export default function login({ client, events }) {
       chunkX: chunk.x,
       chunkZ: chunk.z,
     })
+
+    set_world_border({ client, radius: 10000 })
+    red_screen({ client, warning_blocks: 10000 })
 
     load_chunks(state, { client, events, chunks: [chunk] })
 
--- a/src/login.js
+++ b/src/login.js
@@ -4,8 +4,7 @@ import { write_brand } from './plugin_channels.js'
 import { dimension_codec, overworld } from './world/codec.js'
 import { load_chunks } from './chunk/update.js'
 import { write_title } from './title.js'
-import { set_world_border, red_screen } from './lowlife.js'
-
+import { set_world_border } from './world_border.js'
 export default function login({ client, events }) {
   events.once('state', (state) => {
     const {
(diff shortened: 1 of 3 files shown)


src/login.js
```diff
@@ -4,6 +4,7 @@ import { write_brand } from './plugin_channels.js'
 import { dimension_codec, overworld } from './world/codec.js'
 import { load_chunks } from './chunk/update.js'
 import { write_title } from './title.js'
+import { set_world_border, red_screen } from './lowlife.js'
```
DeltaEvo, line 7: I think since it deals with border this should be named `world_border.js`

src/lowlife.js
```diff
@@ -0,0 +1,8 @@
+export function set_world_border({ client, radius }) {
+  client.write('world_border', { action: 0, radius })
+  client.write('world_border', { action: 2, x: 0, z: 0 })
```
DeltaEvo, line 3: `x` and `z` should be passed as argument

DeltaEvo, line 3: action 0 & 2 can be combined in one action 3 https://wiki.vg/Protocol#World_Border

src/lowlife.js
```diff
@@ -0,0 +1,8 @@
+export function set_world_border({ client, radius }) {
+  client.write('world_border', { action: 0, radius })
```
DeltaEvo, line 2: I think we should define an enum for the actions

src/lowlife.js
```diff
@@ -0,0 +1,8 @@
...
+  client.write('world_border', { action: 2, x: 0, z: 0 })
+}
+
+export function red_screen({ client, warning_blocks }) {
+  client.write('world_border', { action: 5, warning_blocks })
+}
```
DeltaEvo, line 8: If this function is for turning the screen red I don't think that we should pass a warning_blocks to it

src/login.js
```diff
@@ -56,6 +57,9 @@ export default function login({ client, events }) {
       chunkZ: chunk.z,
     })
 
+    set_world_border({ client, radius: 10000 })
```
DeltaEvo, line 60: Instead of `10000` we should use the map dimension, `0 0 1020 1020`

DeltaEvo (changes requested):

src/lowlife.js
ToufouMaster: an exemple pls

src/login.js
ToufouMaster: 1020/2 = 510 radius

src/login.js
ToufouMaster: ok

src/lowlife.js
ToufouMaster: ok

ToufouMaster: ok

src/lowlife.js
ToufouMaster: ok

src/login.js
ToufouMaster: ok

src/lowlife.js
DeltaEvo: https://github.com/aresrpg/aresrpg/blob/master/src/chat.js#L11-L15

src/login.js
```diff
@@ -56,6 +56,8 @@ export default function login({ client, events }) {
...
```
DeltaEvo, line 59: Some parameters seems to be missing `x`, `z`, ...

DeltaEvo (changes requested):

src/login.js
ToufouMaster: oh i see why it does'nt worked

src/login.js
```diff
@@ -56,6 +56,8 @@ export default function login({ client, events }) {
       chunkZ: chunk.z,
     })
 
+    set_world_border({ client, x: 512, z: 510, radius: 10000, speed: 1 })
```
DeltaEvo, line 59: Instead of 10000 we should use the map dimension, `0 0 1020 1020`
And why `x` and `z` are not equals ?

src/world_border.js
```diff
@@ -0,0 +1,36 @@
+export const Actions = {
+  SET_RADIUS: 0, // Set the world border radius
+  LERP_RADIUS: 1, // Set the world border radius in x seconds with a slide animation
+  SET_CENTER: 2, // Set the world border position
+  INIT: 3, // Do all others actions at same times
+  SET_WARNING_TIME: 4, // i don't know (default : 0)
```
DeltaEvo, line 6:
```suggestion
  SET_WARNING_TIME: 4, // (default : 0)
```

ToufouMaster: xD

src/login.js
ToufouMaster: i will fix that sry

ToufouMaster: ok

src/world_border.js
ToufouMaster: ok

DeltaEvo (approved):

PR #99, opened 2021-01-07
Add reaction commands

ToufouMaster:
/rt /tg /thug /ragequit...


src/commands/ragequit.js
```diff
@@ -0,0 +1,23 @@
...
+export default function ragequit({ sender }) {
+  sender.write('chat', {
+    message: JSON.stringify([
+      { text: ' ' + sender.username, color: 'gray' },
+      { text: ' a RageQuit !', color: 'red' },
+    ]),
+    position: Position.CHAT,
+    sender: sender.uuid,
+  })
+  sender.write('kick_disconnect', { reason: JSON.stringify('§ert si c trist') })
```
DeltaEvo, line 22: Use `client.end` here like un ressource_pack.js
And create a json chat instead of using color codes

DeltaEvo (changes requested):

src/commands/f.js
```diff
@@ -0,0 +1,25 @@
...
+  {
+    flags: {
+      command_node_type: 1,
+    },
+    extraNodeData: 'f',
+    children: [],
+  },
+]
+
+export default function f({ sender }) {
```
DeltaEvo, line 13: Maybe name this file and his function fragile instead

ToufouMaster: good idea

src/commands/dislike.js
```diff
@@ -0,0 +1,24 @@
...
+    flags: {
+      command_node_type: 1,
+    },
+    extraNodeData: 'dislike',
+    children: [],
+  },
+]
+
+export default function dislike({ sender }) {
+  sender.write('chat', {
```
DeltaEvo, line 14: Using that the chat message is only sent to the player that have issued the command, use the function inside chat.js to send it to all the players

ToufouMaster: hum i remember having an issue with chat.js function but Okay i will try that

src/commands/ragequit.js
ToufouMaster: Ok

src/commands/f.js
ToufouMaster: Ok

src/commands/dislike.js
ToufouMaster: Ok

DeltaEvo (approved):

PR #128, opened 2021-02-14
Behavior Trees

DeltaEvo:


src/debug.js
```diff
@@ -0,0 +1,109 @@
...
+    stream: new PassThrough({ objectMode: true }),
+  }))
+
+  app.get('/behavior/:id', (request, reply) => {
+    reply.type('text/event-stream')
+    reply.header('Connection', 'keep-alive')
+
+    function format(data) {
+      return `data: ${JSON.stringify(data)}\n\n`
+    }
```
Sceat, line 46: why a verbose func and not a const ?
```js
const format = data => `data: ${JSON.stringify(data)}\n\n`
```

Sceat (approved):

PR #132, opened 2021-02-17
Exploring core integration with enjin

Sceat:


src/index.js
```diff
@@ -140,7 +142,11 @@ aiter(on(server, 'login')).reduce(
 
     const events = new EventEmitter()
 
-    aiter(combineAsyncIterators(actions, packets))
+    aiter(
+      combineAsyncIterators(
+        ...[actions[Symbol.asyncIterator](), packets, Enjin.sources]
```
DeltaEvo, line 147:
```suggestion
        actions[Symbol.asyncIterator](), packets, ...Enjin.sources
```

Sceat:
Enjin should be used only as 
- a way to fetch additional inventory for players
- a additional database to store certain type of items in case the user wallet is linked

Sceat:
@DeltaEvo in https://github.com/aresrpg/aresrpg/commit/f90dff7a1ca12d9fec127f3af828df301060c451 i'm trying to create an enjin user by checking in the state if we already created it before, thing is a i can't access the state as it returns `null` not sure how i should design this

DeltaEvo:
> @DeltaEvo in https://github.com/aresrpg/aresrpg/commit/f90dff7a1ca12d9fec127f3af828df301060c451 i'm trying to create an enjin user by checking in the state if we already created it before, thing is a i can't access the state as it returns `null` not sure how i should design this

You need to wait for the first state event, look in login.js

PR #143, opened 2021-02-21
Traders look at players

ToufouMaster:


src/trade/look_player.js
```diff
@@ -0,0 +1,43 @@
+import { chunk_index, chunk_position } from '../chunk.js'
+
+export function look_player({ client, world, events }) {
+  events.on('state', ({ position }) => {
+    const { x, z } = position
```
DeltaEvo, line 5:
```suggestion
  events.on('state', ({ position: { x, z } }) => {
```

src/trade/look_player.js
```diff
@@ -0,0 +1,43 @@
...
+      chunk_position(x) + 1,
+    ]
+    const z_chunks = [
+      chunk_position(z) - 1,
+      chunk_position(z),
+      chunk_position(z) + 1,
+    ]
+    for (const i of x_chunks) {
+      for (const j of z_chunks) {
+        for (const k in traders_in_chunk.get(chunk_index(i, j))) {
```
DeltaEvo, line 19: use `for ... of` here

src/trade/look_player.js
```diff
@@ -0,0 +1,43 @@
...
+          const entity = traders_in_chunk.get(chunk_index(i, j))[k]
+          const yaw = Math.min(
+            127,
+            Math.max(
+              -128,
+              -((Math.atan2(x - entity.x, z - entity.z) / Math.PI / 2) * 255)
+            )
+          )
```
DeltaEvo, line 27:
```suggestion
          const yaw = Math.floor((-Math.atan2(x - entity.x, z - entity.z) / Math.PI) * (255 / 2))
```
Should do the same since atan2 return a value between -PI and PI

src/trade/look_player.js
```diff
@@ -0,0 +1,43 @@
...
+          const target = entity.id
```
DeltaEvo, line 28: `entity_id` or `entityId` seems to be a better name than `target`

src/trade/look_player.js
```diff
@@ -0,0 +1,43 @@
...
+          client.write('entity_head_rotation', {
+            entityId: target,
+            headYaw: yaw,
+          })
+          client.write('entity_look', {
+            entityId: target,
+            yaw,
+            pitch: 0,
+            onGround: true,
+          })
```
DeltaEvo, line 38: Do you need to send both ?
What is the effect of only the entity_head_rotation ?
And what is the one of entity_look ?

DeltaEvo (changes requested):
Some small changes and a question

src/trade/look_player.js
ToufouMaster: doesn't work

DeltaEvo: Why ? What doesn't work ?

DeltaEvo: For example we do it here https://github.com/aresrpg/aresrpg/blob/master/src/trade/spawn_villagers.js#L41-L43

ToufouMaster: ` for (const k in traders_in_chunk.get(chunk_index(i, j))) {`

i don't know why but i can't do this cause it return undefined

DeltaEvo: What returns undefined ?

ToufouMaster: ohh i understand now

ToufouMaster: i didn't divide it into multiple variable

DeltaEvo: You don't need to
```js
for (const entity in traders_in_chunk.get(chunk_index(i, j))) {
}
```
Works as well

DeltaEvo: We are just destructuring it in spawn_villagers

src/trade/look_player.js
ToufouMaster: done

src/trade/look_player.js
ToufouMaster: done

src/trade/look_player.js
ToufouMaster: done

src/trade/look_player.js
ToufouMaster: done

src/trade/look_player.js
ToufouMaster: done

src/trade/look_player.js
```diff
@@ -0,0 +1,33 @@
...
+    ]
+    const z_chunks = [
+      chunk_position(z) - 1,
+      chunk_position(z),
+      chunk_position(z) + 1,
+    ]
+    for (const i of x_chunks) {
+      for (const j of z_chunks) {
+        if (traders_in_chunk.has(chunk_index(i, j))) {
+          for (const k of traders_in_chunk.get(chunk_index(i, j))) {
```
DeltaEvo, line 19: rename k to `trader`

DeltaEvo (approved):

PR #154, opened 2021-02-27
Use event for chat

Zekirax:
use event and not server.clients


package-lock.json
```diff
@@ -1,4460 +1,8 @@
 {
   "name": "@aresrpg/aresrpg",
   "version": "1.0.0",
-  "lockfileVersion": 2,
```
DeltaEvo, line 4: Can you remove theses changes ?

DeltaEvo (changes requested):
Noice, with that you can also remove the TODO here https://github.com/aresrpg/aresrpg/blob/master/src/index.js#L110

DeltaEvo (approved):

PR #156, opened 2021-02-28
Add msg command

Zekirax:


src/commands/msg.js
```diff
@@ -0,0 +1,55 @@
...
+
+  const private_message = args.slice(1).join(' ')
+
+  const options = {
+    message: JSON.stringify({ text: private_message }),
+    position: Position.CHAT,
+    sender: sender.uuid,
+  }
+
+  world.events.emit('private message', { receiver_username: args[0], options }) // args[0] is the username
```
DeltaEvo, line 54: Maybe create a `username` variable instead of `args[0] is the username`

src/commands/msg.js
```diff
@@ -0,0 +1,55 @@
...
+              name: 'text',
+              parser: 'brigadier:string',
+              properties: 2,
+            },
+          },
+        ],
+        extraNodeData: {
+          name: 'player',
+          parser: 'minecraft:entity',
+          properties: 0x02, // 0x02 for only player entity
```
DeltaEvo, line 33: Can you create a constant for that ?

src/commands/msg.js
```diff
@@ -0,0 +1,55 @@
...
+    extraNodeData: 'msg',
+    children: [
+      {
+        flags: {
+          command_node_type: 2,
+        },
+        children: [
+          {
+            flags: {
+              command_node_type: 2,
```
DeltaEvo, line 19: Maybe we should create a constant for that

src/player/chat.js
```diff
@@ -87,5 +87,8 @@ export default {
       )
     })
     world.events.on('chat', (options) => client.write('chat', options))
+    world.events.on('private message', ({ receiver_username, options }) => {
```
DeltaEvo, line 90: Try to avoid using spaces in event names, `private_message` is better

DeltaEvo (changes requested):

src/commands/msg.js
Zekirax: what is "that"? Be more precise

src/commands/msg.js
Zekirax: what is "that"? Be more precise

src/commands/msg.js
DeltaEvo: that, is the line on which I made the comment, but to be more precise, create a constant for the magic number `0x2`

src/commands/msg.js
DeltaEvo: The `2` constant for the command_node_type

src/commands/declare_options.js
```diff
@@ -0,0 +1,17 @@
...
+    ALL: 0x01, // If set, only allows a single entity/player
+    PLAYER: 0x02, // If set, only allows players
+  },
+}
+
+export const command_node_types = {
+  ROOT_INDEX: 0,
+  COMMAND: 1,
+  ARGUMENT: 2,
+}
```
DeltaEvo, line 17: Can you name that with CamelCase so `parser_properties` => `ParserProperties` and `command_node_types` => `CommandNodeTypes`

DeltaEvo (changes requested):
Nice :+1:, just two small little changes :)

src/commands/msg.js
```diff
@@ -0,0 +1,58 @@
...
+  },
+]
+
+export default function msg({ world, sender, args }) {
+  if (args.length < 2) {
+    write_error({ sender })
+    return
+  }
+
+  const [username, ...message_words] = args
```
DeltaEvo, line 47: Perfect :+1:

Zekirax: I know. I'm the best.

DeltaEvo (approved):

PR #198, opened 2021-05-07
Chicken behavior

DeltaEvo:


package.json
```diff
@@ -39,6 +39,7 @@
     "pino": "^6.11.1",
     "prismarine-provider-anvil": "^2.3.1",
     "typescript": "^4.1.5",
+    "vec3": "^0.1.7",
```
loucass003, line 42: We didn't had that already ? with i used that in the screen feature.

src/behavior/random_position.js
```diff
@@ -0,0 +1,33 @@
...
+    path_position({
+      path: state.path,
+      start_time: state.start_time,
+      speed: state.speed,
+      time,
+    })
+  )
+
+  const x_delta = Math.floor(Math.random() * 15 - 7)
+  const z_delta = Math.floor(Math.random() * 15 - 7)
```
loucass003, line 17: this can gives non-air block ?

package.json
DeltaEvo: Yep but I just list it as a direct dependency

src/behavior/random_position.js
DeltaEvo: This can give us invalid positions but it's not an issue for now since the pathfinding will just fail, in the future we should try to only take valid positions

DeltaEvo:
`Fix stack overflow of recursive async iterators` seems to not fix the issue

loucass003 (approved):

PR #211, opened 2021-06-15
Added bot system

NEXOmega:
**TODO :** 

- Re-Enable Ressource Pack, disabled because bot would not accept it and stay blocked before joining server
- Finish Command to start test individually


package.json
```diff
@@ -36,9 +36,13 @@
     "iterator-helper": "^1.1.3",
     "minecraft-data": "^2.74.0",
     "minecraft-protocol": "^1.23.1",
+    "mineflayer": "^3.7.0",
+    "mineflayer-move": "github:cheezbarger/mineflayer-move",
```
DeltaEvo, line 40: This should be moved to devDependencies

DeltaEvo (changes requested):
Nice ! Thanks for taking the time to implement that, but I don't think it should be part of the "main" code we should separate the tests from the code,
also I was thinking to use https://jestjs.io/ for the tests

Can you try to separate what you did from the main code (inside a `test` folder instead of `src`) ?

Sceat:
> Nice ! Thanks for taking the time to implement that, but I don't think it should be part of the "main" code we should separate the tests from the code,
> also I was thinking to use [jestjs.io](https://jestjs.io/) for the tests
> 
> Can you try to separate what you did from the main code (inside a `test` folder instead of `src`) ?

We could also use https://github.com/HydreIO/doubt

package.json
```diff
@@ -6,6 +6,7 @@
   "type": "module",
   "scripts": {
     "start": "node src/index.js | pino-human",
+    "test": "node --experimental-vm-modules ./node_modules/jest/bin/jest.js",
```
Sceat, line 9: Let's discuss before futher implementation as i actually tried Jest and it sucks with esmodules, really don't think we should use it, @hydre/doubt or node-tap are great.

src/index.js
```diff
@@ -143,8 +144,8 @@ const mobs_position = mobs_position_factory(world)
 async function observe_client(context) {
   /* Observers that handle the protocol part.
    * They get the client and should map it to minecraft protocol */
-
-  await player_resource_pack.observe(context)
+  if (!context.client.username.startsWith('#iRobot'))
+    await player_resource_pack.observe(context)
```
Sceat, line 148: probably @DeltaEvo will know how to handle this

tests/bot.js
```diff
@@ -0,0 +1,7 @@
+import mineflayer from 'mineflayer'
```
Sceat, line 1: the folder name should be `test`

tests/connected.test.js
```diff
@@ -0,0 +1,17 @@
...
+  return new Promise((resolve) => {
+    bot.on('spawn', () => {
+      resolve(true)
+    })
+  })
+}
+
+test('Can connect', () => {
+  return connected().then((data) => {
+    expect(data).toBeTruthy()
```
Sceat, line 15: this is great! i like the idea of simply implementing a testing base

tests/connected.test.js
```diff
@@ -0,0 +1,17 @@
...
+
+import bot from './bot.js'
+
+function connected() {
+  return new Promise((resolve) => {
+    bot.on('spawn', () => {
+      resolve(true)
+    })
+  })
+}
```
Sceat, line 11:
```suggestion
import { expect, test } from '@jest/globals'
import { once } from 'events'

import bot from './bot.js'

const connected = () => events.once(bot, 'spawn')
```

Sceat (changes requested):
Thanks again 🎉

Sceat:
closing for inactivity

PR #234, opened 2021-08-02
Add a attack_target behavior node

DeltaEvo:


src/mobs/attack.js
```diff
@@ -0,0 +1,25 @@
...
+
+import { aiter } from 'iterator-helper'
+
+import { Types } from './types.js'
+
+export default {
+  /** @type {import('../index.js').Observer} */
+  observe({ client, world, dispatch }) {
+    for (const mob of world.mobs.all) {
+      aiter(on(mob.events, 'state')).reduce(
```
Sceat, line 11: might need to update with abortable

src/player/health.js
```diff
@@ -2,7 +2,26 @@ import { on } from 'events'
...
+
+      log.info({ damage, health }, 'Damage')
+
+      return {
+        ...state,
+        health,
+      }
+    }
+    return state
+  },
```
Sceat, line 24: this already exist in the `master` branch :thinking:

Sceat (commented):
What's the state of this PR ?

src/behavior/attack_target.js
```diff
@@ -0,0 +1,23 @@
+import { SUCCESS, RUNNING, FAILURE } from '../behavior.js'
+
+export default function attack_target(node, state) {
+  if (state.target == null) {
+    return {
+      status: FAILURE,
+      state,
+    }
+  } else if (state.target_position == null) {
```
Sceat, line 9: i don't see where you are actually using the difference between `null` and `undefined` cuz if it's unused we should stick to only one representation of emptyness for simplicity 🤔

Sceat (changes requested):

Sceat:
u have a conflict

src/behavior/attack_target.js
DeltaEvo: I am not using that `== null` (not the double equal) will match for null and undefined

Sceat: i assume that if you don't use `(!state.target)` it's because you want to differentiate `null` and `undefined` so that `undefined` means it was just created and `null` means it lost his target, but if you don't use such behaviour then why using `null` at all

Sceat (approved):

PR #250, opened 2021-08-08
Refactor the gamemode command

Sceat:


src/commands/gamemode.js
```diff
@@ -10,6 +10,14 @@ const GameMode = {
   SPECTATOR: 3,
 }
 
+const is_gamemode_valid = (number) => number >= 0 && number <= 3
```
DeltaEvo, line 13:
```suggestion
const is_gamemode_value = (value) => Object.values(GameMode).includes(value)
```

src/commands/gamemode.js
```diff
@@ -10,6 +10,14 @@ const GameMode = {
...
+const parse_gamemode = (param) => {
+  if (is_gamemode_valid(+param)) return +param
+  return GameMode[param.toUpperCase()]
+}
+const mode_name_from_number = (number) =>
```
DeltaEvo, line 18:
```suggestion
const gamemode_from_value = (value) =>
```

DeltaEvo (changes requested):

src/commands/gamemode.js
```diff
@@ -54,35 +62,33 @@ export const gamemode_nodes = [
...
+          with: [
+            {
+              translate: `gameMode.${gamemode_from_value(
+                gameMode
+              ).toLowerCase()}`,
+            },
+          ],
+        }),
+        position: Position.CHAT,
+        sender: sender.uuid,
```
DeltaEvo, line 87: Keep the null UUID here, if you want to avoid a magic value introduce a constant with the name `SERVER_UUID` in index.js or in this file

DeltaEvo (approved):

PR #262, opened 2021-08-16
Implement a permanent storage

Sceat:
Here are some ideas of a structure we could use, please review and let's discuss.


DeltaEvo:
Some early thoughts:
- I don't think we have the need for redis graph, we won't have a lot of relations (I don't think we will have relations at all), I think a bare redis will do the work and be easier we could switch later if needed
- Loading a player from database should not use the classic observer/reducer struct instead it should provide the initial_state of the player

```javascript
let state = initial_state
if (USE_PERSISTANT_STORAGE)
  state = load_from_database(initial_state, client.uuid)
```

src/database/graph.js
```diff
@@ -0,0 +1,25 @@
...
+    get(_, prop) {
+      return Reflect.get(graph, prop)
+    },
+  }
+}
+
+// i swear i'm going to punch your face with your damn types
+// how the fuck you'd type a proxy
+/** @type {any} */
+export default new Proxy({}, handler)
```
DeltaEvo, line 25: Why are you using a proxy for that, it seems a bit overkill for me, can you not just create yhe "noop" object yourself ?

src/database/graph.js
```diff
@@ -0,0 +1,25 @@
...
+const { USE_STORAGE = 'false' } = process.env
+const GRAPH_NAME = 'aresrpg'
+let handler = {
+  get() {
+    return () => ({})
+  },
+}
+
+// not sure about this, but pretty sure delta will know what to do
+if (USE_STORAGE?.toLowerCase() === 'true') {
```
DeltaEvo, line 12: You can move USE_STORAGE into settings.js and do the convertion to a boolean here

Sceat:
I'm afraid redis is a bit limited as a database, basic stuff like secondary key indexing needs us to do more work, redisearch is good but still limited and more overhead to learn that redisgraph (and still not a database), look at the doc for Disk :/ the choice for me is either mongodb or redisgraph, while mongodb force us to express relations with foreign keys

DeltaEvo:
> I'm afraid redis is a bit limited as a database, basic stuff like secondary key indexing needs us to do more work, redisearch is good but still limited and more overhead to learn that redisgraph (and still not a database), look at the doc for Disk :/ the choice for me is either mongodb or redisgraph, while mongodb force us to express relations with foreign keys

The thing is that for me we will not need to use the complex parts of redisearch, for the game data we only need a Map<UUID, JSON>, so no secondary key, we don't even need a structure since we will only access the data as a whole, I think we should just store the json as a string first in redis, then if we need to acces part of the data use redisjson (maybe the website will need that)

Sceat:
Forgot the existence of redisjson 👍 that's indeed the best and simplest choice for us

Sceat:
PR is ready :+1: yet there is an infinite amount of typing error, deal with it

src/database/sentinel.js
```diff
@@ -0,0 +1,57 @@
...
+const master_client = new Redis(redis_options('master'))
+
+await Promise.all([
+  events.once(slave_client, 'ready'),
+  events.once(master_client, 'ready'),
+])
+
+log.info({ master: true, slave: true }, 'redis cluster initialized')
+
+new Set([master_client, slave_client]).forEach((client) => {
```
DeltaEvo, line 53: You don't need to create a set here

src/database/sentinel.js
```diff
@@ -0,0 +1,57 @@
...
+  client.on('error', () => {})
```
DeltaEvo, line 54: Don't ignore error of clients, log them or let them be unhandled

src/database/sentinel.js
```diff
@@ -0,0 +1,57 @@
...
+
+const slave_client = new Redis(redis_options('slave'))
+const master_client = new Redis(redis_options('master'))
+
+await Promise.all([
+  events.once(slave_client, 'ready'),
+  events.once(master_client, 'ready'),
+])
+
+log.info({ master: true, slave: true }, 'redis cluster initialized')
```
DeltaEvo, line 51: Why do we need to have `{ master: true, slave: true }` if it's constant ?

src/database/sentinel.js
```diff
@@ -0,0 +1,57 @@
...
+
+const {
+  REDIS_HOST = '0.0.0.0',
+  REDIS_SENTINEL_PORT = 26379,
+  REDIS_MASTER_NAME = 'mymaster',
+} = process.env
+
+const log = logger(import.meta)
+
+const probe_status = {
```
DeltaEvo, line 15: Why do we need to probe the status ?

DeltaEvo (changes requested):

src/database/index.js
```diff
@@ -0,0 +1,43 @@
...
+    position,
+    view_distance,
+    inventory,
+    game_mode,
+    experience,
+    health,
+  })
+
+export default (client) => ({
+  push: async (state) => {
```
DeltaEvo, line 27:
```suggestion
  async push(state) {
```

src/database/index.js
```diff
@@ -0,0 +1,43 @@
...
+    return Clients.write.call(
+      'JSON.SET',
+      client.uuid.toLowerCase(),
+      '.',
+      serialize(state)
+    )
+
+    // we should also trigger a redis subscription here to sync all nodes
+  },
+  pull: async () => {
```
DeltaEvo, line 39:
```suggestion
  async pull() {
```

src/database/sentinel.js
```diff
@@ -0,0 +1,57 @@
...
+  status: true,
+}
+
+const retryStrategy = (label) => (attempt) => {
```
DeltaEvo, line 19: snake_case ?
I would prefer that you use functions instead of array functions for top level functions

src/database/sentinel.js
```diff
@@ -0,0 +1,57 @@
...
+  events.once(master_client, 'ready'),
+])
+
+log.info({ master: true, slave: true }, 'redis cluster initialized')
+
+new Set([master_client, slave_client]).forEach((client) => {
+  client.on('error', () => {})
+})
+
+export { master_client, slave_client, probe_status }
```
DeltaEvo, line 57: named exports ?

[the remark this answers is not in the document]
Sceat: redis sentinel is an alternative to a cluster, since i already had a compact code from Sidy i just used it
read more here https://stackoverflow.com/questions/31143072/redis-sentinel-vs-clustering

src/database/sentinel.js
Sceat: it makes the module kubernetes ready

src/database/sentinel.js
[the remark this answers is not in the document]
DeltaEvo: Is it really ? I just read the documentation of redis, so I might be wrong, but from what I understood sentinel only monitor and manage the cluster

But even with that information I don't think we should "handle" the shape of the redis deployment in our code, the library should handle that

src/database/sentinel.js
DeltaEvo: Can you explain why ?

src/index.js
```diff
@@ -208,12 +214,19 @@ function create_context(client) {
...
+  const player_state = await storage.pull()
   const controller = new AbortController()
-  client.once('end', () => {
+
+  client.once('end', async () => {
     log.info(
       { username: client.username, uuid: client.uuid },
       'Client disconnected'
     )
+    await storage.push(get_state())
```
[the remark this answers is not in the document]
Sceat: how do i get that ?

src/database/sentinel.js
Sceat: if the redis connection is lost, it allows the k8s scheduler to not move load to the current pod, we'll have to simply implement a socket/http response to whatever route we configure in k8s (like `/health`)

Should we terminate the process if there are players on this instance ? the drop could last just the time for k8s to reboot redis, Ioredis has a queue to replay commands that hapenned while it was reconnecting.

There could be a flow in this, if all pods can't contact redis well they may all be restarted i'm don't remember well how the scheduler respond to probes status but this discussion should be detailed when we host stuff.

src/database/index.js
Sceat: what's the difference ?

src/database/sentinel.js
Sceat: does that change anything ?

src/database/sentinel.js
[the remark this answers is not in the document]
Sceat: it designate new masters and manage replicas but that's different from the official "redis-cluster" image.
Would be better to not handle it but that's very specific with redis, you only have one client in normal deployments where you read and write, with sentinel you have a master and many slaves to read from. You also can't move from a simple redis to a clustered redis without dumping and importing again with the cluster system in mind. Sentinel is kinda an enhanced version of the simple client and that's the best we can use without needing to deploy a minimum of 6 instance (which requires a more heavy and expensive k8s)
image

So this "shape selection" is actually a constant (we may never need a cluster unless we reach ten of thousands of players) without this file we can't use sentinel, or we should write an external module exporting a read and write client (overkill ?)

src/database/sentinel.js
Sceat: I don't but that was an eslint rule (don't remember which one)

src/index.js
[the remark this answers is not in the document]
DeltaEvo: The reduce here https://github.com/aresrpg/aresrpg/blob/master/src/index.js#L236 return a promise that will resolve with the last state value

src/database/sentinel.js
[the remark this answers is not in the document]
DeltaEvo: Can we not create one client and use the read-write splitting of ioredis ?

For the sentinel configuration I think we should make it optional and change the docker compose to be dead simple and only have a rejson master and only use the sentinel option on the production cluster

src/database/sentinel.js
DeltaEvo: Ok but currently we terminate the process directly, so the probe_status will never be used, I think we should remove it since we don't know yet how we will implement that and a boolean flag might not be enough

package.json
```diff
@@ -21,38 +21,47 @@
...
   "author": "AresRPG's Team",
   "license": "MIT",
   "bugs": {
     "url": "https://github.com/aresrpg/aresrpg/issues"
   },
   "homepage": "https://github.com/aresrpg/aresrpg#readme",
   "dependencies": {
-    "@aresrpg/aresrpg-map-colors": "^1.1.3",
-    "canvas": "^2.6.1",
+    "@aresrpg/aresrpg-map-colors": "^1.1.2",
```
DeltaEvo, line 36: Don't update the deps in this PR

src/database/index.js
DeltaEvo: Follow the current style of the codebase and I don't see the need of an arrow function here, it's not an anonymous function and you don't need to use the lexical `this`

src/database/sentinel.js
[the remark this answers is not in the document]
Sceat: ioredis support read/write for clusters only, we can switch back to a single redis node but we'll have to get sentinel pretty quickly anyway (or switch directly to a cluster ?)

package.json
```diff
@@ -21,18 +21,25 @@
...
+  },
   "author": "AresRPG's Team",
   "license": "MIT",
   "bugs": {
     "url": "https://github.com/aresrpg/aresrpg/issues"
   },
   "homepage": "https://github.com/aresrpg/aresrpg#readme",
   "dependencies": {
-    "@aresrpg/aresrpg-map-colors": "^1.1.3",
+    "@aresrpg/aresrpg-map-colors": "^1.1.2",
```
DeltaEvo, line 36: Mhh it seems that you are downgrading a dependency :thinking:

src/index.js
```diff
@@ -228,16 +248,37 @@ function create_context(client) {
...
+      .then((state) =>
+        Redis.push({
+          key: client.uuid.toLowerCase(),
+          value: serialize_state(state),
+        })
+      )
+      .catch((error) => {
+        // TODO: what to do here if we can't save the client ?
+        log.error(error)
+      })
```
DeltaEvo, line 279: You don't need to do that in the `once('end'` :smile: 
This mean that you can avoid introducing a local variable and chain the `.then(state)` after the reduce

src/index.js
```diff
@@ -208,14 +232,10 @@ function create_context(client) {
     'Client connected'
   )
 
+  /** @type {NodeJS.EventEmitter} */
+  const events = new EventEmitter()
```
DeltaEvo, line 236: Can you reput that line where it was ? (Before the aiter)

src/database/sentinel.js
DeltaEvo: I unresolved this since this is still in camelCase

src/database/sentinel.js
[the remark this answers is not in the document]
DeltaEvo: I don't see why read/write splitting would not work with sentinel :thinking:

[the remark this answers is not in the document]
DeltaEvo: Ah got it it's on a `Redis.Cluster` object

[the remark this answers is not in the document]
DeltaEvo: https://github.com/luin/ioredis/blob/master/lib/redis/RedisOptions.ts#L9-L11 RedisOptions extends the IClusterOptions so you can use `scaleReads` on a bare redis client :)

package.json
Sceat: dang wtf i copy paster master

src/database/sentinel.js
Sceat: oooops

src/database/sentinel.js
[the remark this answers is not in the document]
DeltaEvo: we can give it but it's not used :(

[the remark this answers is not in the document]
Sceat: get rekt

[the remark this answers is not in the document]
DeltaEvo: I will resolve that since I don't see how to do better then :\

src/index.js
```diff
@@ -209,6 +233,8 @@ function create_context(client) {
   )
 
   const controller = new AbortController()
+
+  client.on('error', (error) => log.error(error, 'Client error'))
```
DeltaEvo, line 237: Can you reput that where it was ?

Sceat: xD hey!!

DeltaEvo (approved):

PR #264, opened 2021-08-20
Add support for held item and off-hand swap

Sceat:


src/index.js
```diff
@@ -107,6 +107,7 @@ const initial_state = {
   inventory_sequence_number: 0,
   inventory_cursor: null,
   inventory_cursor_index: 0,
+  inventory_held_slot_index: 0, // goes from 0 to 8
```
DeltaEvo, line 110: Don't want to describe that with a type :smile:  ?
```suggestion
  /** @type {0|1|2|3|4|5|6|7|8} */ 
  inventory_held_slot_index: 0,
```

src/player/inventory.js
```diff
@@ -25,50 +25,84 @@ const BlockDigStatus = {
   SWAP_ITEM_IN_HAND: 6,
 }
 
+const OFF_HAND_SLOT = 45
```
DeltaEvo, line 28: Can you move your code into a new file instead, like src/player/held_item.js ?

src/player/inventory.js
```diff
@@ -25,50 +25,84 @@ const BlockDigStatus = {
...
+      state.inventory_held_slot_index
+    )
+    ;[state.inventory[OFF_HAND_SLOT], state.inventory[held_item_slot]] = [
+      state.inventory.at(held_item_slot),
+      state.inventory.at(OFF_HAND_SLOT),
+    ]
+    return {
+      ...state,
+      inventory: state.inventory,
+      inventory_sequence_number: state.inventory_sequence_number + 1,
```
DeltaEvo, line 82: We don't need to resync the inventory here

src/player/inventory.js
```diff
@@ -25,50 +25,84 @@ const BlockDigStatus = {
...
+
+  // swap second hand
+  if (status === 6) {
+    const held_item_slot = hotbar_to_inventory_slot(
+      state.inventory_held_slot_index
+    )
+    ;[state.inventory[OFF_HAND_SLOT], state.inventory[held_item_slot]] = [
+      state.inventory.at(held_item_slot),
+      state.inventory.at(OFF_HAND_SLOT),
+    ]
```
DeltaEvo, line 78: Don't mutate the state
Don't use at if you don't have negative indexes

src/index.js
DeltaEvo: I think we should remove the `inventory_` prefix here

DeltaEvo (changes requested):

src/index.js
Sceat: > Don't want to describe that with a type 😄 ?

image

DeltaEvo:
Mmh it seems that you removed the code that was writing the `held_item_slot`

Sceat:
fixed ✔️ 
and actually we needed the inventory sequence incrementation

DeltaEvo (approved):

PR #273, opened 2021-08-21
Disable parenthesis in prettier

Sceat:


DeltaEvo:
Mmh did you read why this was the default on prettier https://prettier.io/docs/en/options.html#arrow-function-parentheses ?

I don't really care here, I just want to be sure that we acknowledge the reason of why it's the default for prettier

Sceat:
> At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option.

Ok it actually make sense, yet seeing parenthesis there trigger in me some desire of pukingm and cleaning a keyboard is tedious

DeltaEvo (approved):

PR #276, opened 2021-08-22
bumping dependencies except husky which sucks in @7

Sceat:


DeltaEvo:
I would prefer to merge the dependabot PRs after landing #281

PR #288, opened 2021-08-27
Players can see each other

### LINKED ISSUE #256: Players can't see other players
Hello we are a multiplayer server but when i connect 2 accounts i'm still alone, is it some kind of reflection of my own life ?

DeltaEvo:
Fix #256

### DIFF
--- a/src/player/sync.js
+++ b/src/player/sync.js
@@ -1,0 +1,169 @@
+import { on } from 'events'
+import { PassThrough } from 'stream'
+
+import { aiter } from 'iterator-helper'
+
+import { is_inside } from '../math.js'
+import { chunk_position, chunk_index, same_chunk } from '../chunk.js'
+import { abortable } from '../iterator.js'
+
+function insert(array, value) {
+  const nullIdx = array.indexOf(null)
+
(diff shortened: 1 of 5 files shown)


src/player/sync.js
```diff
@@ -0,0 +1,169 @@
...
+
+  if (nullIdx === -1) {
+    return { array: [...array, value], index: array.length }
+  } else {
+    return {
+      array: [...array.slice(0, nullIdx), value, ...array.slice(nullIdx + 1)],
+      index: nullIdx,
+    }
+  }
+}
```
Sceat, line 21: why not using a Set instead of manipulating arrays ?

Sceat (approved):

src/player/sync.js
DeltaEvo: Because I need stable indexes to use as entity id

Sceat: should'nt we rethink this id management ? random instead of increments

DeltaEvo: We could, but I don't think it will be simpler :thinking:

PR #274, opened 2021-08-21
Add support for basic scoreboards

Sceat:
- removed unused fields from ares v1 (stuff never supported)
- polished design

image

PR #290, opened 2021-08-29
Add support for basic scoreboards

Sceat:
reopened (force pushed), tracked from #274


src/player/scoreboard.js
```diff
@@ -0,0 +1,120 @@
...
+const PROGRESS_SQUARES_AMOUNT = 12
+const KARES_FORMATER = Intl.NumberFormat('en', { notation: 'compact' })
+
+const Slots = {
+  EMPTY_GROUP_SLOT: { color: 'gray', text: '-' },
+  CLASS: ({ name, level, progress }) => [
+    { text: `${name}`, color: 'white', bold: true },
+    { text: ' [Lvl ', color: 'gray' },
+    { text: `${level}`, color: 'dark_green', reset: false },
+    { text: '] (', color: 'gray' },
```
DeltaEvo, line 25: I'm not a big fan of repeating the experience on the scoreboard, we already have a dedicated bar for that

src/scoreboard/inline_component.js
```diff
@@ -0,0 +1,50 @@
...
+
+  return `${Colors[color?.toUpperCase()] ?? ''}${inline_modifiers}${text}${extra
+    .map(extract_modifiers)
+    .join('')}`
+}
+
+// sending chat components is great but it's adding extra reset chars
+// between all components parts (even extra)
+// which increase way too much the length of the scoreboard line
+// makes the client crash, so we need to convert back to string
```
DeltaEvo, line 48: It's fixed no ?

src/scoreboard/inline_component.js
```diff
@@ -0,0 +1,50 @@
...
+      if (modifiers[key.toLowerCase()]) return `${combined}${value}`
+      return combined
+    },
+    ''
+  )
+
+  return `${Colors[color?.toUpperCase()] ?? ''}${inline_modifiers}${text}${extra
+    .map(extract_modifiers)
+    .join('')}`
+}
```
DeltaEvo, line 43: `extract_modifier` doesn't really mean what this function does, `component_to_legacy` seems a better name for me

src/scoreboard/inline_component.js
```diff
@@ -0,0 +1,50 @@
+import normalize from './normalize.js'
```
DeltaEvo, line 1: This file should go into a src/chat.js instead

DeltaEvo: And normalize should be inlined in it

src/scoreboard/inline_component.js
```diff
@@ -0,0 +1,50 @@
...
+    .map(extract_modifiers)
+    .join('')}`
+}
+
+// sending chat components is great but it's adding extra reset chars
+// between all components parts (even extra)
+// which increase way too much the length of the scoreboard line
+// makes the client crash, so we need to convert back to string
+export default raw_component =>
+  normalize(raw_component).map(extract_modifiers).join('')
```
DeltaEvo, line 50: We should name this function `components_to_legacy` seems a good name

src/scoreboard/inline_component.js
```diff
@@ -0,0 +1,50 @@
...
+const extract_modifiers = ({ text, extra = [], color, ...modifiers }) => {
+  const inline_modifiers = Object.entries(Modifiers).reduce(
+    (combined, [key, value]) => {
+      if (modifiers[key.toLowerCase()]) return `${combined}${value}`
+      return combined
+    },
+    ''
+  )
+
+  return `${Colors[color?.toUpperCase()] ?? ''}${inline_modifiers}${text}${extra
```
DeltaEvo, line 40: You should call the default function (components_to_legacy) on `extra`

src/scoreboard/inline_component.js
DeltaEvo: Or inline that into `extract_modifiers`

src/scoreboard/update_sidebar.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export default function ({ client, scoreboard_name }) {
+  return ({ last, next }) =>
+    next
+      .map(normalize)
+      .map((component, index) => ({ component, index }))
+      .filter(only_changes(last))
+      .map(no_duplicates(next))
+      .map(create_packets(scoreboard_name, last))
```
DeltaEvo, line 65: flatMap ?

src/scoreboard/update_sidebar.js
```diff
@@ -0,0 +1,67 @@
+import equals from 'fast-deep-equal'
+
+import inline_component from './inline_component.js'
+import normalize from './normalize.js'
+
+const MAGIC_RESET = '§r'
```
DeltaEvo, line 6: This should go into `src/chat.js`

DeltaEvo (changes requested):

src/player/scoreboard.js
Sceat: let's open a separate issue on restyling the scoreboard because right now without this line the overall look would be less cool, i think that in the end there will be no scoreboard but simply a datapack

Sceat: image

PR #307, opened 2021-09-18
Speed up server port bind speed

### LINKED ISSUE #306: Speed up start time
Aresrpg is already being too slow to boot up, let's isolate and make stuff optional or lazy loaded, like arborist

DeltaEvo:
To do that we dynamic import the whole state/context logic
to let the server initialize first

The server now bind to port in 500ms, the bigger slowdown
now is inside minecraft-protocol: it's the generation of
the rsa key

Fix #306


Sceat (approved):
lool this dude is not joking 🚀

Issue #325, opened 2021-10-02
Show new level title


Sceat:
- Indicate to a player when he just passed a level (in title)
- Indicate to nearby players when a player just passed a level (100 block ?) (in chat)

PR #330, opened 2021-10-18
Visual responses from gaining XP and Levels

Sceat:


Sceat:
only thing missing for #325 is to send a msg to nearby players, but even looking at #288 i'm not sure yet how to do that

DeltaEvo:
For that you need to send an event on the world with as payload the chat message a position (and maybe a radius), listen this event and the position of the player is within the radius, send the message

Sceat:
this is also missing https://hastebin.com/epomujevuz.java

src/context.js
```diff
@@ -115,6 +119,11 @@ const initial_state = {
   game_mode: 2,
   experience: 0,
   health: 40,
+  max_health: 40,
+  soul: 100,
+  statistics: {
+    available_points: 0,
+  },
```
DeltaEvo, line 126: I don't think we need to store the available_points since you can compute that with the level and the used points

src/context.js
```diff
@@ -115,6 +119,11 @@ const initial_state = {
   game_mode: 2,
   experience: 0,
   health: 40,
+  max_health: 40,
```
DeltaEvo, line 122: I don't think we need to store the max_health since this can be computed

src/player/experience.js
```diff
@@ -109,59 +118,261 @@ const levels = [
...
-  const [current_level, current_level_experience] = reversed_levels.find(
-    ([level, level_experience]) => level_experience <= total_experience
+  const [current_level] = reversed_levels.find(
+    ([level, level_experience]) => total_experience >= level_experience
   )
-  return {
-    level: current_level,
-    remaining_experience: total_experience - current_level_experience,
-  }
+  return current_level
```
DeltaEvo, line 140: Why did you changed that ?

src/player/experience.js
```diff
@@ -109,59 +118,261 @@ const levels = [
...
-  const next_level_experience = levels[level + 1]
-  return (
-    remaining_experience / (next_level_experience - current_level_experience)
-  )
+export function level_progress({ current_level, total_experience }) {
+  const current_level_experience = levels[current_level] ?? Infinity
+  const next_level_experience = levels[current_level + 1]
+  const current_level_progression = total_experience - current_level_experience
+  const required_experience = next_level_experience - current_level_experience
+  return Math.round((100 * current_level_progression) / required_experience)
```
DeltaEvo, line 151: I don't think the convertions to percents should be donne here

src/player/experience.js
```diff
@@ -109,59 +118,261 @@ const levels = [
...
-  if (level + 1 >= levels.length) {
-    return 0
-  }
-  const next_level_experience = levels[level + 1]
-  return (
-    remaining_experience / (next_level_experience - current_level_experience)
-  )
+export function level_progress({ current_level, total_experience }) {
+  const current_level_experience = levels[current_level] ?? Infinity
+  const next_level_experience = levels[current_level + 1]
```
DeltaEvo, line 148: You don't handle the case where `current_level` is the last level and there is no level after that

src/player/experience.js
```diff
@@ -109,59 +118,261 @@ const levels = [
...
 export default {
+  /** @type {import('../context.js').Reducer} */
+  reduce(state, { type, payload }) {
+    if (type === 'player/add_experience') {
+      const { experience } = payload
+      return {
+        ...state,
+        experience: state.experience + experience,
+      }
+    } else if (type === 'player/new_level') {
```
DeltaEvo, line 163: We should be able to detect a new level directly when we have a add_experience

src/player/experience.js
```diff
@@ -109,59 +118,261 @@ const levels = [
...
+            }
+          }
+          return {
+            last_total_experience: total_experience,
+            last_level: current_level,
+          }
+        },
+        {
+          last_total_experience: 0,
+          last_level: 1,
```
DeltaEvo, line 374: This initial state doesn't seems right, what if the player is not level 1 🤔

src/player/experience.js
```diff
@@ -109,59 +118,261 @@ const levels = [
...
+              write_sound({
+                client,
+                sound: SOUND.EXPERIENCE_ORB,
+                ...position,
+              })
+            }
+          }
+          return {
+            last_total_experience: total_experience,
+            last_level: current_level,
```
DeltaEvo, line 369: Do we really need to store the last_level here, the last_total_experience doesn't already have this information ?

DeltaEvo (changes requested):

src/player/experience.js
Sceat: Code wasn't working properly (also ambiguous)

Sceat:
Oii this review will need to make my brain do maths and i'm bad at it 😵

DeltaEvo (approved):

PR #337, opened 2021-11-01
Add missing loot data

DeltaEvo:


Sceat:
We should discuss about hidding loots and chances in a database for gameplay reasons

Sceat (approved):

PR #355, opened 2021-11-10
Implement simplified soul system

Sceat:
- [x] reduce soul on death
- [x] at 0% the player become a ghost (being a ghost and canceling interraction is outside the scope of this PR)
- [x] regenerate soul over time (except in ghost mode)


src/context.js
```diff
@@ -115,6 +119,13 @@ const initial_state = {
   game_mode: 2,
   experience: 0,
   health: 40,
+  body_state: BodyStates.ALIVE,
```
DeltaEvo, line 122: I don't think we need to store the `body_state` in the state since it can be derived from the soul and health, I feel this is redundant data

src/context.js
```diff
@@ -321,8 +337,12 @@ export async function create_context(client) {
         events.emit(Context.STATE, state)
         return state
       },
-      { ...initial_state, ...player_state }
+      { ...initial_state, ...player_state, username: client.username }
```
DeltaEvo, line 340: I don't think the username should be part of the state. we can maybe give a third argument to the reducer with some "context" and this could include the username

src/player/login.js
```diff
@@ -117,6 +129,9 @@ export default {
         },
         canvas
       )
+
+      // make sure all pods are running with UTC time
+      dispatch(Action.LOGIN, { time: Date.now() })
```
DeltaEvo, line 134: I think this should be part of the initial state

src/player/soul.js
```diff
@@ -0,0 +1,79 @@
...
+
+    aiter(abortable(on(events, Context.STATE, { signal })))
+      .map(([{ soul }]) => soul)
+      .reduce((last_soul, soul) => {
+        if (last_soul !== soul && soul === 0)
+          dispatch(Action.BODY_STATE, { body_state: BodyStates.GHOST })
+        return soul
+      })
+
+    events.once(
```
DeltaEvo, line 64: I don't think you have the good `last_connection_time` here, for me it will arrive in the second state update

DeltaEvo (changes requested):
The DEAD state seems a little weird since it's not a state that we stay in 🤔

Sceat:
> The DEAD state seems a little weird since it's not a state that we stay in 🤔 

We do while we're in the respawn countdown, and there different interactions may be restricted

src/context.js
Sceat: hum that might get complex without it,
- we can't increase soul when someone is a ghost
- a player can be dead and have health
- a player can be a ghost and have health 
- a player can be alive and have no soul ? 
there is many possible case

src/context.js
Sceat: could we have a nickname maybe ? thats a basic feature

src/context.js
DeltaEvo: > a player can be dead and have health
Mmh, this seems weird, can you explain how does that happen ?

> a player can be a ghost and have health
Yes we should consider a player to be a Ghost when he have no sould

> a player can be alive and have no soul ?
Seems weird too 🤔, the player loose sould when he die right ?

For me we can have (in this order of priority):
- soul == 0 -> Ghost
- health == 0 -> Dead
- -> Alive

src/context.js
DeltaEvo: Mmh, indeed a nickname would fit here, but I don't think we should use the nickname in the logs 😅, the thing that bother me here is to have a username that is constant inside the state

src/player/login.js
Sceat: wow shit jeez that's smart

src/commands/health.js
```diff
@@ -0,0 +1,36 @@
...
+export default function health({ world, sender, dispatch, args }) {
+  if (args.length === 1) {
+    const [health] = args
+    if (!Number.isNaN(+health)) {
+      dispatch(Action.HEALTH, { health: +health })
+      client_chat_msg({
+        client: sender,
+        message: [
+          { text: 'health updated: ', color: '#ECF0F1', italic: true },
+          { text: health, color: '#E74C3C', bold: true, italic: false },
```
DeltaEvo, line 29: This should use the new colors no ?

src/player/soul.js
```diff
@@ -0,0 +1,67 @@
...
+
+const HOUR_1 = 3_600_000
+const MINUTE_10 = 600_000
+const SOUL_REGEN_PER_OFFLINE_HOUR = 5
+const SOUL_REGEN_PER_ONLINE_HOUR = 12
+
+export default {
+  /** @type {import('../context.js').Reducer} */
+  reduce(state, { type, payload }) {
+    if (type === Action.DEATH) {
```
DeltaEvo, line 19: I don't think you need to have a dispatch here, just check if the health is at 0, and move the reduce from the player sould after the one for the player health in context.js

DeltaEvo: Mmh, forget this comment it's not a good idea 😅, we are not able to see if it got to 0 in the current update

DeltaEvo (approved):

src/context.js
```diff
@@ -115,6 +117,12 @@ const initial_state = {
   game_mode: 2,
   experience: 0,
   health: 40,
+  // player's energy, losing after each death
+  soul: 100,
+  // last time the player joined,
+  // can be used for example to calcule regenerated soul while offline
+  last_connection_time: Date.now(),
```
DeltaEvo, line 124: This is not right, this mean that everyone will have the same last_connection_time, you need to set that at line 343

DeltaEvo (changes requested):

src/commands/health.js
Sceat: Indeed, the PR was not merged at that time

DeltaEvo (approved):

PR #359, opened 2021-11-14
Improve entity update performance and remove screens

Sceat:


DeltaEvo:
Can you split this PR in two or remove the `divide by 2 the amount of entity updates` ?

Sceat:
> Can you split this PR in two or remove the `divide by 2 the amount of entity updates` ?

What's the problem if we don't squash ? will it not be 2 distinct updates in the history ?

DeltaEvo:
Because this is two different update, I'm fine to merge the remove screen spawning but not the divide by two of entity updates

DeltaEvo (approved):

PR #361, opened 2021-11-14
Prevent a player in creative mode from taking damages

Sceat:


src/player/damage.js
```diff
@@ -20,10 +24,32 @@ export function register(world) {
   }
 }
 
+const kUndefined = Symbol('bruh')
+// base ground for future extensions (ex: stats)
+const Damages = state => ({
+  fall: ({ damage }) => damage,
+  // great pattern matching /shrug
+  [kUndefined]: () => {
```
DeltaEvo, line 32: `({ [undefined]: 1 })[undefined]` or  `({ [null]: 1 })[null]` works too :)

src/player/damage.js
```diff
@@ -20,10 +24,32 @@ export function register(world) {
...
+    throw new Error('no damange source found')
+  },
+})
+
```
DeltaEvo, line 36: I don't really see the need for that, I prefer that we add it we we will have more than one damage source

DeltaEvo (changes requested):

src/player/damage.js
Sceat: I replace the `FALL_DAMAGE` action by a generic `DAMAGE` one so still need pattern matching here, that small addition is quite mild

DeltaEvo (approved):

PR #363, opened 2021-11-14
Send an action bar to the player

Sceat:
image

image

### DIFF
--- a/src/chat.js
+++ b/src/chat.js
@@ -2,7 +2,7 @@ import { World } from './events.js'
 import { SERVER_UUID } from './settings.js'
 
 /**
- * Take a string or a chat component as an input and output a chat component
+ * Take anything as an input and output an array of chat component
  */
 export function normalize_chat_component(component) {
   if (!component) return [{ text: '' }]
--- a/src/context.js
+++ b/src/context.js
@@ -24,6 +24,7 @@ import player_gamemode from './player/gamemode.js'
 import player_scoreboard from './player/scoreboard.js'
 import player_block_place from './player/block_placement.js'
 import player_bossbar from './player/boss_bar.js'
+import player_action_bar from './player/action_bar.js'
 import player_traders, {
   register as register_player_traders,
 } from './player/traders.js'
@@ -125,6 +126,7 @@ const initial_state = {
   health: 40,
   // player's energy, losing after each death
   soul: 100,
+
   // last time the player joined,
   // can be used for example to calcule regenerated soul while offline
   last_connection_time: undefined,
@@ -256,6 +258,7 @@ export async function observe_client(context) {
   player_teleportation_stones.observe(context)
   player_tablist.observe(context)
   player_sync.observe(context)
+  player_action_bar.observe(context)
   player_scoreboard.observe(context)
   player_block_place.observe(context)
   player_item_loot.observe(context)
--- a/src/player/action_bar.js
+++ b/src/player/action_bar.js
@@ -1,0 +1,111 @@
+import { on } from 'events'
+import { setInterval } from 'timers/promises'
+
+import { aiter } from 'iterator-helper'
+
+import { Formats } from '../chat.js'
+import { Context } from '../events.js'
+import { abortable } from '../iterator.js'
+import {
+  get_max_health,
+  get_remaining_stats_point,
+} from '../player_statistics.js'
+import { write_action_bar } from '../title.js'
+
+function to_rgb(percent) {
+  if (percent < 50)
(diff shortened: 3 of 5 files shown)


src/player/action_bar.js
```diff
@@ -0,0 +1,99 @@
...
+
+function to_rgb(percent) {
+  if (percent < 50)
+    return { red: 255, green: Math.round(5.1 * percent), blue: 0 }
+  return { red: Math.round(510 - 5.1 * percent), green: 255, blue: 0 }
+}
+
+function to_hex({ red, green, blue }) {
+  const hue = red * 0x10000 + green * 0x100 + blue * 0x1
+  return `#${('000000' + hue.toString(16)).slice(-6)}`
```
DeltaEvo, line 18: `padStart` instead of the `'000000'` + `slice(-6)`

src/player/action_bar.js
```diff
@@ -0,0 +1,99 @@
...
+      ...compute_stats_component(remaining_stats_point),
+      { text: ' <<', ...Formats.BASE, italic: false, bold: true },
+    ],
+  })
+}
+
+export default {
+  /** @type {import('../context.js').Observer} */
+  observe({ client, get_state, world, events, signal }) {
+    aiter(abortable(setInterval(2000))).forEach(() => {
```
DeltaEvo, line 61: You need to pass the signal to the setInterval, to be able to cancel it when the player disconnect

DeltaEvo: Also why are you updating it each 2 seconds nothing seems to be time based here ?

src/player/action_bar.js
```diff
@@ -0,0 +1,99 @@
...
+        health,
+        max_health: get_max_health(),
+        remaining_stats_point: get_remaining_stats_point(),
+      })
+    })
+
+    aiter(abortable(on(events, Context.STATE, { signal })))
+      .map(([{ health, get_remaining_stats_point, get_max_health }]) => ({
+        health,
+        // might be costly :/
```
DeltaEvo, line 74: Why ?

src/chat.js
```diff
@@ -73,6 +73,14 @@ export function client_chat_msg({ client, message }) {
   })
 }
 
+export function client_action_bar({ client, message }) {
+  client.write('chat', {
```
DeltaEvo, line 77: There is also a dedicated action bar packet, I think we should use it

src/player/action_bar.js
```diff
@@ -0,0 +1,99 @@
...
+        remaining_stats_point: get_remaining_stats_point(),
+        max_health: get_max_health(),
+      }))
+      .reduce(
+        (
+          [last_health, last_max_health, last_remaining_stats_point],
```
DeltaEvo, line 80: Why this is an array and not an object ?

DeltaEvo (changes requested):

src/player/action_bar.js
Sceat: the action bar disappears on the client

DeltaEvo: Ah crap, even with the dedicated action bar packet ?

Sceat: the what ?

DeltaEvo: There is another packet to send an action bar
https://wiki.vg/Protocol#Action_Bar maybe it will not do that with this one

src/context.js
```diff
@@ -121,6 +122,11 @@ const initial_state = {
   health: 40,
   // player's energy, losing after each death
   soul: 100,
+
+  // those 2 will be computed according to @deltaevo
+  get_remaining_stats_point: () => 5,
+  get_max_health: () => 40,
```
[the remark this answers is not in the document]
Sceat: about the max_health i can now take it from the `types` (#234) but for the stats point i'm waiting your update on #330 to know how to get it

src/player/action_bar.js
Sceat: depends on how you implement the stats point computation

src/player/action_bar.js
Sceat: it still disapear with that packet

Sceat: also timings will only affect real titles

src/context.js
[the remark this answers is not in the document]
Sceat: see [`dd22a61` (#363)](https://github.com/aresrpg/aresrpg/pull/363/commits/dd22a61e680cb9feb5b98678bd26722a38249082)

DeltaEvo (approved):

PR #366, opened 2021-11-24
Send Targeted Entity Life with a Boss Bar

Sceat:


src/boss_bar.js
```diff
@@ -0,0 +1,51 @@
+import uuid from 'uuid-1345'
+
+export const Actions = {
```
DeltaEvo, line 3: Can you keep the comments please ?

src/player/boss_bar.js
```diff
@@ -1,89 +1,100 @@
...
- * @property {string} text - Required
- * @property {boolean} [bold]
- * @property {boolean} [italic]
- * @property {boolean} [underlined]
- * @property {boolean} [strikethrough]
- * @property {boolean} [obfuscated]
- * @property {string} [color]
- * @property {Array<Title>} [extra]
- */
+const UUID_LENGTH = 36
```
DeltaEvo, line 11: `UUID_V4_LENGTH`

src/player/boss_bar.js
```diff
@@ -1,89 +1,100 @@
...
+      noDisposeOnSet: true,
+      dispose(entityUUID) {
+        write_bossbar({
+          client,
+          entityUUID,
+          action: Actions.REMOVE,
+        })
+      },
+      maxAge: 6000,
+    })
```
DeltaEvo, line 46: A LRU cache for that seems very overkill, I think a simple ringbuffer of 3 elements would do the job here like for the damange indicators

src/player/boss_bar.js
```diff
@@ -1,89 +1,100 @@
...
+    events.on(Context.MOB_DAMAGE, ({ mob, damage }) => {
+      const {
+        displayName,
+        max_health = 20 /* i assume max_health will be there */,
+      } = Types[mob.mob]
+      const { health } = mob.get_state()
+      const entityUUID = `${mob.entity_id}`.padEnd(
+        UUID_LENGTH - `${mob.entity_id}`.length,
+        '0'
+      )
```
DeltaEvo, line 60: I don't think this makes a valid uuid, a valid uuid should have dashes in it like `b7084b42-51c0-4132-8bdc-b66a74e4b19d`

DeltaEvo (changes requested):

src/player/boss_bar.js
Sceat: but it works 😵

DeltaEvo: I will check the code of prismarine to see if it rrally works then, but I don't think we should rely on that

Sceat: we already rely on something like this for the `SERVER_UUID`

DeltaEvo: This doesn't mean it's a good thing, but yeah we migh change that one too

src/player/boss_bar.js
Sceat: 😢

src/player/boss_bar.js
Sceat: prismarine seems to format those as Pakkit show me correct uuids

DeltaEvo (approved):

Issue #365, opened 2021-11-24
Player screen throw an error when any click is made


Sceat:
```
context  E  State error {
  stack: "TypeError: Cannot read properties of undefined (reading 'x')\n" +
    '    at Vec3.dot (/Users/sceat/dev/aresrpg/aresrpg/node_modules/vec3/index.js:126:27)\n' +
    '    at screen_ray_intersection (file:///Users/sceat/dev/aresrpg/aresrpg/src/player/screen.js:187:15)\n' +
    '    at EventEmitter.<anonymous> (file:///Users/sceat/dev/aresrpg/aresrpg/src/player/screen.js:220:29)\n' +
    '    at Object.onceWrapper (node:events:514:26)\n' +
    '    at EventEmitter.emit (node:events:406:35)\n' +
    '    at aiter.map.reduce (file:///Users/sceat/dev/aresrpg/aresrpg/src/context.js:337:16)\n' +
    '    at HAsyncIterator.reduce (/Users/sceat/dev/aresrpg/aresrpg/node_modules/iterator-helper/dist/async-iterator/HAsyncIterator.js:208:25)\n' +
    '    at runMicrotasks (<anonymous>)\n' +
    '    at processTicksAndRejections (node:internal/process/task_queues:96:5)',
  type: 'Error'
}
```

PR #367, opened 2021-11-25
Fix player screen throw an error when any click is made

Sceat:
✅ Closes: #365

### DIFF
--- a/src/context.js
+++ b/src/context.js
@@ -72,6 +72,7 @@ const initial_world = {
   next_window_id: 1, // 0 is the player inventory
   /** @type {() => Object} Remove type to remove circular references */
   get: () => world,
+  screens: {},
 }
 
 const world_reducers = [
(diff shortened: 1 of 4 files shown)


Sceat:
@DeltaEvo i can't understand the pipeline 🤯

src/context.js
```diff
@@ -245,7 +245,7 @@ export async function observe_client(context) {
   player_health.observe(context)
   player_attributes.observe(context)
   player_chat.observe(context)
-  player_screen.observe(context)
+  // player_screen.observe(context)
```
DeltaEvo, line 248: You don't need to remove that

src/context.js
```diff
@@ -81,10 +81,10 @@ const world_reducers = [
...
   register_player_traders,
   register_player_deal_damage,
-  register_screen({
-    id: 'player_screen',
-    size: { width: 8, height: 4 },
-  }),
+  // register_screen({
+  //   id: 'player_screen',
+  //   size: { width: 8, height: 4 },
+  // }),
```
DeltaEvo, line 87: Remove it instead of comment it

DeltaEvo (changes requested):

Sceat:
resolved but what the heck is wrong with the pipeline 🤯

DeltaEvo:
What do you mean by pipeline ?

Sceat:
the CI

DeltaEvo (approved):

Issue #364, opened 2021-11-24
Killing a mob with no loots throws an error


Sceat:
```
TypeError: Types[mob.mob].loots is not iterable
    at EventEmitter.<anonymous> (file:///Users/sceat/dev/aresrpg/aresrpg/src/mobs/loot.js:14:55)
    at EventEmitter.emit (node:events:394:28)
    at file:///Users/sceat/dev/aresrpg/aresrpg/src/mobs/damage.js:61:22
    at HAsyncIterator.reduce (/Users/sceat/dev/aresrpg/aresrpg/node_modules/iterator-helper/dist/async-iterator/HAsyncIterator.js:208:25)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Sceat:
like killing a npc #339

Sceat:
see #369

PR #369, opened 2021-11-25
Killing a mob with no loots now works

Sceat:
✅ Closes: #364

### DIFF
--- a/src/mobs/loot.js
+++ b/src/mobs/loot.js
@@ -11,7 +11,8 @@ export default {
       const owner = mob.get_state().first_damager
 
       if (owner === client.uuid) {
-        for (const { chance, item } of Types[mob.mob].loots) {
+        const { loots = [] } = Types[mob.mob]
+        for (const { chance, item } of loots) {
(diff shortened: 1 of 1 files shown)


DeltaEvo:
I don't think this is a valid fix, a mob should always have a loot array
The one that currently don't have one are:
- hades_horse: Not spawned because hades is not implemented
- hades: Not implemented
- numen_citizen: A NPC, should not be killable
- segeste_citizen: A NPC, should not be killable
- thebes_citizen: A NPC, should not be killable
- guardian: A NPC, should not be killable
- monk: A NPC, should not be killable

Sceat:
wouldn't it be the same to add a loot array in the mob ? this PR just look safer

DeltaEvo:
Yes it would be the same but I'm not asking that, for example for the NPCs I don't wan't to add a loot array because they should not have loots.
This should not be an issue because an NPC should not die
If it die I prefer them server to crash, than for it to continue silently
This commit will just mask the error

Sceat:
then it's fixed with #368

Issue #151, opened 2021-02-26
Show a msg when a player join for the first time


Sceat:
![image](https://user-images.githubusercontent.com/11330271/109317874-7a9f3080-784d-11eb-9e36-637745970408.png)

ToufouMaster:
it requires a database right?

Sceat:
currently our database is just a state object in memory

DeltaEvo:
For now we can just add a first_time in the state that is always true

ToufouMaster:
k

Sceat:
> it requires a database right?

supported in #262

PR #370, opened 2021-11-25
Show a msg when a player join for the first time

Sceat:
✅ Closes: #151


Sceat:
note that i had to change the signature of `world_chat_msg` before using it for the feature!

DeltaEvo (approved):

PR #378, opened 2021-11-29
Add support for entity metadata with a json file

Sceat:
@DeltaEvo i just don't know how to handle bitmasks if you can commit directly on the PR to fix that 🙏 

- both the way to write them in the json (https://wiki.vg/Entity_metadata#Entity_Metadata_Format)
- and the `Bitmask()` function

### DIFF
--- a/src/entity_metadata.js
+++ b/src/entity_metadata.js
@@ -1,0 +1,66 @@
+import entity_metadata from './entity_metadata.json'
+
+export const villager_types = {
+  DESERT: 0,
+  JUNGLE: 1,
+  PLAINS: 2,
+  SAVANNA: 3,
+  SNOW: 4,
+  SWAMP: 5,
+  TAIGA: 6,
+}
+
+export const villager_professions = {
+  NONE: 0,
+  ARMORE: 1,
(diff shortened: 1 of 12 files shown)


src/entity_metadata.js
```diff
@@ -0,0 +1,59 @@
...
+function to_bitmask(bitflags, object) {
+  return Object.entries(bitflags).reduce(
+    (result, [name, value]) => result | (object[name] << value),
+    0
+  )
+}
+
+export function to_metadata(entity, metadatas) {
+  const { parent, metadata } = entity_metadata[entity] ?? {}
+  if (!metadata) throw new Error(`${entity} doesn't have any metadata`)
```
DeltaEvo, line 40: An entity can have a parent and no metadata for example the silverfish https://wiki.vg/Entity_metadata#Silverfish
```suggestion
  if (!(entity in entity_metadata)) throw new Error(`${entity} doesn't have any metadata`)
  const { parent, metadata } = entity_metadata[entity]
```

DeltaEvo (changes requested):

DeltaEvo (approved):

PR #402, opened 2021-12-08
Cleaner implementation of Bossbar, damage indicator, and damage invulnerability

Sceat:
I isolated away the ring buffer and the lru so the code become clearer (in my pov). Commit history is cleaned

### DIFF
--- a/data/entities.json
+++ b/data/entities.json
@@ -1,8 +1,8 @@
 {
   "talus_pig": {
     "displayName": "Porc de Talus",
-    "type": "mob",
-    "mob": "pig",
+    "category": "mob",
+    "minecraft_entity": "pig",
     "xp": 10,
     "loots": [
       {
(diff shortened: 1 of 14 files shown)


src/lib/lru_set.js
```diff
@@ -0,0 +1,59 @@
+const no_key =
+  key =>
+  ({ key: filtered_key } = { key: undefined }) =>
+    filtered_key !== key
+
+export default function lru_set({ capacity, on_eviction, max_age }) {
+  let cache = Array.from({ length: capacity })
```
DeltaEvo, line 7: I'm not a big fan of having a mutable datastructure in a code where we try to avoid mutability where possible

Sceat: Should i externalize this module ? this is low level enough and pretty well isolated to not care too much about it

Sceat: https://www.npmjs.com/package/@hydre/lru_set

DeltaEvo (approved):

PR #415, opened 2022-01-05
Fix dependabot codeowner syntax

Sceat:
Seems CODEOWNERS are broken by github, nothing we can do about it


DeltaEvo:
https://github.com/github/feedback/discussions/8959

PR #426, opened 2022-01-05
Implement death countdown

Sceat:


Sceat:
need #402

src/player/respawn.js
```diff
@@ -0,0 +1,115 @@
...
+            entityId: client.uuid,
+            effectId: BLINDNESS,
+            amplifier: 1,
+            duration: respawn_in * 20,
+            hideParticles: true,
+          })
+
+          aiter(
+            abortable(setInterval(1000, null, { signal: controller.signal }))
+          ).reduce(remaining => {
```
DeltaEvo, line 57: I think you can avoid using a custom signal with a https://github.com/tc39/proposal-iterator-helpers#takelimit

DeltaEvo (changes requested):

src/player/respawn.js
Sceat: yo that's clever

DeltaEvo (approved):

Issue #153, opened 2021-02-26
Show player health below his name


Sceat:
![image](https://user-images.githubusercontent.com/11330271/109320499-598c0f00-7850-11eb-9da0-80658cfcd667.png)

PR #430, opened 2022-01-18
Show player health below his name

Sceat:
closes: #153


Sceat:
doesn't support sync, see #431

DeltaEvo (approved):

Issue #429, opened 2022-01-18
State is null when 2 players join at the same time


Sceat:
`player/tablist.js`

The position cannot be found because the state is `null`, we may need to make that get_state asynchronous

```js
    const on_player = info => {
      // Add player to tab list
      world.events.emit(World.ADD_PLAYER(client.uuid), info)
      if (info.UUID !== client.uuid) {
        // Send my info to new player
        world.events.emit(
          World.ADD_PLAYER(info.UUID),
          player_info(get_state().position)
        )
      }
    }
```

DeltaEvo:
I think we should instead register the on_player in the once(Context.STATE, ...)

PR #432, opened 2022-01-18
fix: 🐛 State is null when 2 players join at the same time

Sceat:
✅ Closes: #429

### DIFF
--- a/src/player/tablist.js
+++ b/src/player/tablist.js
@@ -17,7 +17,6 @@ export default {
         data: [info],
       })
     }
-    world.events.on(World.ADD_PLAYER(client.uuid), add_player)
 
     const on_player = info => {
       // Add player to tab list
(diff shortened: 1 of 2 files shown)


src/player/tablist.js
```diff
@@ -30,14 +30,14 @@ export default {
...
     }
-    world.events.on(World.PLAYER, on_player)
 
     client.once('end', () => {
       world.events.off(World.ADD_PLAYER(client.uuid), add_player)
       world.events.off(World.PLAYER, on_player)
     })
 
     events.once(Context.STATE, ({ position }) => {
+      world.events.on(World.PLAYER, on_player)
```
DeltaEvo, line 40: We should also move the `world.events.on(World.ADD_PLAYER(client.uuid), add_player)`

DeltaEvo (approved):

Issue #105, opened 2021-01-16
Experience when you kill a mob


DeltaEvo:
Depends on #13 

https://docs.google.com/document/d/1wXd8ZTcBAEc25vP8Tx5JtV37An6Hwuk43tCahPABnEc/edit?usp=sharing

Old code

```java
/*     */   public static void calcAndAddPlayerExp(Player p, int exp, int mobLvl) {
/*  90 */     byte maxLVL = 100;
/*  91 */     if (GroupeManager.hasGroupe(p)) {
/*  92 */       ArrayList lvl = GroupeManager.getGroupeMembers(p);
/*  93 */       ArrayList<Player> message = new ArrayList();
/*  94 */       Iterator<String> var7 = lvl.iterator();
/*     */       
/*  96 */       while (var7.hasNext()) {
/*  97 */         String loc = var7.next();
/*  98 */         Player lvl1 = Bukkit.getPlayerExact(loc);
/*  99 */         if (lvl1 != null && lvl1.isOnline() && lvl1.getLocation().distance(p.getLocation()) <= 50.0D) {
/* 100 */           message.add(lvl1);
/*     */         }
/*     */       } 
/*     */       
/* 104 */       if (message.size() != 0) {
/* 105 */         if (message.size() > 1) {
/* 106 */           exp = (int)Math.round(exp * 1.25D);
/*     */         }
/*     */         
/* 109 */         exp = (int)Math.round(exp / message.size());
/* 110 */         if (exp <= 0) {
/* 111 */           exp = 1;
/*     */         }
/*     */         
/* 114 */         var7 = (Iterator)message.iterator();
/*     */         
/* 116 */         while (var7.hasNext()) {
/* 117 */           Player loc3 = (Player)var7.next();
/* 118 */           if (!loc3.isDead()) {
/* 119 */             int lvl3 = loc3.getLevel();
/* 120 */             int expp = getExpForLevel(lvl3, mobLvl, exp);
/* 121 */             if (StatsStorage.getPlayerAme(loc3) <= 40) {
/* 122 */               expp = (int)(expp * 0.5D);
/*     */             }
/*     */             
/* 125 */             if (expp != 0) {
/* 126 */               if (lvl3 < maxLVL) {
/* 127 */                 ExpStorage.addPlayerExp(loc3, expp);
/*     */               }
/*     */               
[...]

PR #489, opened 2022-07-10
xp loot

SohKey:
- Add xp loot from killing mobs
- Add visible + ? xp on mob kill
image


src/mobs/loot.js
```diff
@@ -1,14 +1,24 @@
...
-  observe({ client, events, dispatch }) {
+  observe({ client, events, world, dispatch }) {
     events.on(Context.MOB_DEATH, ({ mob }) => {
       const position = mob.position()
       const owner = mob.get_state().first_damager
 
+      dispatch(Action.ADD_EXPERIENCE, { experience: +Entities[mob.type].xp })
+
+      const { damage_indicator_start_id } = world
+      const entity_id = damage_indicator_start_id
```
Sceat, line 16: you can't use the damage indicator id, it should be a new set of ids for experience, or generalize the indicators ids

src/mobs/loot.js
```diff
@@ -1,14 +1,24 @@
...
+      create_armor_stand(client, entity_id, position, {
+        text: `+${Entities[mob.type].xp}xp`,
```
Sceat, line 18: this was already called before, it's better to not duplicate code and make use of variables

Sceat (commented):
Thanks for the contribution, there are few problems concerning this PR

- You added the code from teleportation stone on this PR, it should not be there.
- All of this should be discussed before any implementation is made, there is an issue #105 which speak of groups and more complex systems including mobs levels against users levels
- experience related code should be in `src/player/experience.js`
- the name of this PR should be more descriptive
- the history of commits should be clean

SohKey:
> Thanks for the contribution, there are few problems concerning this PR
> 
> * You added the code from teleportation stone on this PR, it should not be there.
> * All of this should be discussed before any implementation is made, there is an issue #105 which speak of groups and more complex systems including mobs levels against users levels
> * experience related code should be in `src/player/experience.js`
> * the name of this PR should be more descriptive
> * the history of commits should be clean

Sorry for the forgots things (history of commits / code from tp stones),

- I moved the experience related code into the right file.
- For the #105 i just wanted code the xp gain for the player only. About groups @DeltaEvo explained me that each player in the group will receive an event that says that the group has gained xp, but this will come after what im doing (just implementing xp gain from player alone).

src/mobs/loot.js
SohKey: It's fix, i will publish a commit later

src/mobs/loot.js
SohKey: looking for to solve this but i think i will need some help to do someting clean.

Sceat:
The point was not to have onlu one commit but a meaningful one, like `feat: dispatch experience event on mob death`

src/player/experience.js
```diff
@@ -357,5 +359,18 @@ export default {
...
     )
+    events.on(Context.MOB_DEATH, ({ mob }) => {
+      const position = mob.position()
+      const { xp } = Entities[mob.type]
+
+      dispatch(Action.ADD_EXPERIENCE, { experience: +xp })
+
+      const { mob_death_id } = world
+      const entity_id = mob_death_id
+      create_armor_stand(client, entity_id, position, {
```
Sceat, line 370:
```suggestion
      create_armor_stand(client, mob_death_id, position, {
```

Sceat:
@DeltaEvo did you supervise the hologram id for this PR ? what if we kill multiple mobs at the same time ? it may need more than one id

DeltaEvo:
No I didn't, I need to discuss with @SohKey for that

PR #492, opened 2022-07-14
fix #252 : EventEmitter memory leak

### LINKED ISSUE #252: Possible EventEmitter memory leak detected. 11 state listeners added to [EventEmitter] ...
We need to figure out what to do with that, do we increase slightly the limit or set it to Infinity
https://nodejs.org/api/events.html#events_emitter_setmaxlisteners_n

Gretorm:
Set defaultMaxListeners to infinity. May be risky?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 import fs from 'fs'
 import { join, dirname } from 'path'
 import { fileURLToPath } from 'url'
+import { EventEmitter } from 'events'
 
 import pino from 'pino'
 import protocol from 'minecraft-protocol'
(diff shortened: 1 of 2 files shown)


DeltaEvo:
> Set defaultMaxListeners to infinity. May be risky?

Yes I think it's not how we want to solve that as this warning is useful, it's just that the default value doesn't fit for AresRPG and we are not really able to have a good default.

Thinking at it we have two big event listeners where we need to fix that:
- The mob event listener, one event is added for each player that sees the mob + a fixed number of listeners. As this was causing some spam since the warning was printed for each mob before it was set to Infinity. But we might be able to remove it today, as before it was not for "each player that sees the mob" but for "each player" so adding a lot of event listeners.
- The player event listener, one event is added for each "observer" and it should not move much

The solutions that I see for those two are:
- For the mob event listener, we could set the limit dynamically based on the number of players that sees the mob. We can implement that by increasing and decreasing the limit when a player sees the mob
- For the player event listener I think we could set it to Infinity, let the "observe" functions register their event listeners and after some time/ticks reset the limit to the current number of listeners + a fixed number (to allow for some space, 10 seems right)

Sceat (changes requested):
see https://github.com/aresrpg/aresrpg/pull/492#issuecomment-1185335153

Issue #50, opened 2020-11-14
Bank Chest


DeltaEvo:
- Open an double chest inventory when clicking on a ender chest
- Add multiple pages

Sceat:
How would Banks works in a decentralized environment ? like in real life we hold peoples money on a dev wallet ? :grimacing:

Sceat:
> How would Banks works in a decentralized environment ? like in real life we hold peoples money on a dev wallet ?

actually a NFT can't be stored in the player's bank, the only way to transfert to another account would be via a transaction, and in case the characters are also a token, well the player can also simply send his character to a wallet containing the stuff

Sceat:
might be duplicate of #526

Sceat:
This issue is basically about implementing a vanilla inventory with some specificactions:

- For now there will be only one bank chest limited to a vanilla double chest space #526 
- In the future there will be multiple fixed chests, so you can already store theses in the state as
```js
soul: 100,
// use your own bank functions, this is just an example
bank: Array.from({
	length: 1,
	0: Array.from({ length: 54 }) // the first bank chest content
})
```
- Handle vanilla in-chest gestures, stacks & movements (also be proactive and add those to the player inventory)
- Use #110 as an example on how to store items data (PR will come soon)

Sceat:
closed in favor of #526

Issue #333, opened 2021-10-23
Handle merging items stacks in the inventory


DeltaEvo:
Grouping two items stacks of the same type in the inventory is not supported

Sceat:
looks like it still work client side, but when you disconnect and reconnect items will not be splitted anymore

PR #523, opened 2022-10-16
Bank Chest 

SohKey:
- issue: #50


SohKey:
Currently, open double chest after clicking on a ender chest & handle item interactions.

TODO: 
- Add multiple pages
- Bug: weird item display on item move within the chest

SohKey:
Weird item stacking is releated to #333

Sceat:
Overall this PR is ok, the feature is working. However the code can be improved and made more readable, there is often occasions of
```js
a.b.c
a.b.d
get_state().x
get_state().y
```
which should be replaced by
```js
const { b: { c, d } } = a
const { x, y } = get_state()
```
There are a lot of layers of `if else` which can be replaced by switchs, mappers, functions.. I didn't looked much beyond that because as we discussed in #526 and with @DeltaEvo, probably we should let vanilla inventory as is, and not do any pagination. A more important issue to clean before this one would be indeed #333

PR #538, opened 2022-11-16
Toufoumaster/first_spell

ToufouMaster:


Sceat:
Can you describe in details what this PR bring to the server and add some screenshots ?
- the branch is out of date
- there is a command that shouldn't exist

PR #545, opened 2022-11-22
Toufoumaster/archery

ToufouMaster:
Need a lot of rework, still a draft


ToufouMaster:
i'm not gonna format the code is prettier this way

DeltaEvo:
The point of having of style guide is to stick to it :)
So format your code :)

ToufouMaster:
it just add a line where there is no need to and make the code ugly, if you really want i'll do it

Sceat:
please don't request reviews for drafts

ToufouMaster:
it is not a draft anymore i formated the code

src/player/archery.js
```diff
@@ -0,0 +1,283 @@
...
+            block: null,
+            position: { x: 255, y: 255, z: 255 },
+          }
+          let predict_pos = wall_predict.position
+          let predict_block = wall_predict.block
+          get_path_collision(world, position, velocity, 60).then(result => {
+            wall_predict = result
+            predict_pos = result.position
+            predict_block = result.block
+          })
```
Sceat, line 114: you are using an async callback to assign values inside a sync reducer, this will result in mutating your variables at unpredictable moments and also destroying all life form on earth, avoid doing that or i call the police. Why even creating 3 variables ??

src/player/archery.js
```diff
@@ -0,0 +1,283 @@
...
+            type: 2,
+            ...position,
+            ...direction_to_yaw_pitch(
+              to_direction(-position.yaw, -position.pitch)
+            ),
+            velocityX: velocity.x,
+            velocityY: velocity.y,
+            velocityZ: velocity.z,
+          })
+          let t = 0
```
Sceat, line 136: `t` ? what is `t`, please use descriptive variables names

src/player/archery.js
```diff
@@ -0,0 +1,283 @@
...
+
+                  client_chat_msg({
+                    client: sender,
+                    message: [
+                      { text: 'Accuracy: ', ...Formats.BASE },
+                      { text: values.accuracy + '%', ...format },
+                      { text: ', Distance: ', ...Formats.BASE },
+                      { text: values.distance + '%', ...Formats.SUCCESS },
+                    ],
+                  })
```
Sceat, line 184: this has nothing to do with shooting arrows

src/player/archery.js
```diff
@@ -0,0 +1,283 @@
...
+                const mob_pos = mob.position()
+                const dist = distance3d_squared(cur_pos, mob_pos)
+                if (dist < 1) {
+                  mob.dispatch(MobAction.DEAL_DAMAGE, {
+                    damage: arrow.damage,
+                    damager: sender.uuid,
+                  })
+                  client.write('entity_destroy', {
+                    entityIds: [arrow.entityId],
+                  })
```
Sceat, line 211: may look cool to let arrows in the mob's body for some time

src/player/archery.js
```diff
@@ -0,0 +1,283 @@
...
+        }
+      )
+
+    events.on(Context.MOB_SPAWNED, ({ mob }) => {
+      const { category } = Entities[mob?.type] ?? {}
+      if (category !== 'npc') visible_mobs[mob.entity_id] = mob
+    })
+    events.on(Context.MOB_DESPAWNED, ({ entity_id }) => {
+      if (entity_id in visible_mobs) delete visible_mobs[entity_id]
+    })
```
Sceat, line 254: handling mob visibility has nothing to do with archery

src/player/archery.js
```diff
@@ -0,0 +1,283 @@
...
+
+    client.on('use_item', ({ hand }) => {
+      if (hand === Hand.MAINHAND) {
+        const { inventory, held_slot_index, position } = get_state()
+        const slot_number = held_slot_index + HOTBAR_OFFSET
+        const item = inventory[slot_number]
+        const state = get_state()
```
Sceat, line 261: please avoid duplicated calls (the state)

Sceat (changes requested):
You have an excellent motivation but unfortunately this feature doesn't even work properly, the code logic is messy and you lack important concepts of the language. Most of the `reduce` body is unreadable and mix multiple different logic. Here is a review mainly on codestyle for you to learn but not even on the core, I suggest you try to handle easier feature as there isn't much to save from this one (in my opinion)

src/player/archery.js
ToufouMaster: woops.

src/player/archery.js
ToufouMaster: forgot to remove... Whoops

src/player/archery.js
ToufouMaster: where can i check for a arrow to collide a specific block outside of this function then?

src/player/archery.js
ToufouMaster: delta told me to use this to get every mob the player can see, it avoid me to check for EVERY possible mobs.

src/player/archery.js
ToufouMaster: > Why even creating 3 variables ??

ah,yes that's completely stupid, apart from that i can't call get_path_collision otherwise, or maybe in the reduce

Sceat:
Not usable & abandonned https://github.com/aresrpg/aresrpg/pull/545#pullrequestreview-1195005695

PR #571, opened 2022-12-03
Add emotes to chat

SohKey:
image

- add emotes to chat mapper
- display rank before name (preview)

the texture pack must be updated to see the emotes ⚠

_Big tanks to Watich for lending me an account (what a bro)_


data/emotes.json
```diff
@@ -0,0 +1,25 @@
+{
+  "providers": [
```
Sceat, line 2: why do you need that additional nesting ?

src/player/chat.js
```diff
@@ -73,6 +74,32 @@ export default {
...
+            },
+          }
+        },
+      }
+    })
+
+    const emotes_mapper = {}
+    emoteFormated.forEach(emote => {
+      Object.assign(emotes_mapper, emote)
+    })
```
Sceat, line 96: The thinking process of creating something in order to fill it right after is imperative programming, in functionnal you would

```js
const emotes_mapper = emotes.providers
	.map(...)
	.reduce(...)
```

src/player/chat.js
```diff
@@ -73,6 +74,32 @@ export default {
...
+
+    const chat_mapper = {
+      ...item_mapper,
+      ...emotes_mapper,
+    }
```
Sceat, line 101: there was previously a nice flat mapper, now it's kinda hidden. Prefer appending your feature to the existing structure instead of moving it.

```
const chat_mapper = {
	'%pos%': () ... etc,
	...emotes.reduce(to_chat_map)
}
```

src/player/chat.js
```diff
@@ -87,20 +114,31 @@ export default {
...
               .split(new RegExp(`(${Object.keys(chat_mapper).join('|')})`))
               .map(part => {
-                for (const pattern in chat_mapper)
+                for (const pattern in chat_mapper) {
                   if (
                     part.match(pattern) &&
                     chat_mapper[pattern](part) !== undefined
                   )
                     return chat_mapper[pattern](part)
+                }
```
Sceat, line 141: this is not a useful addition

src/player/chat.js
```diff
@@ -87,20 +114,31 @@ export default {
...
+              text: ' ',
+              font: 'minecraft:playerprofile',
+            },
             {
               text: client.username,
             },
+            {
+              text: ' >> ',
+              color: 'dark_gray',
+            },
```
Sceat, line 131: I think the `chat.type.text` was important @DeltaEvo , also are you sure about that piece of code ? `' >> '` seem a big change maybe not the most good looking one

Sceat (changes requested):
## About `data/emotes.json`
Take a look at `aresrpg/ares-data` it is a new submodule where these have been moved, it is yet to be reviewed by @DeltaEvo but if it's merged the correct place will be in `aresrpg/ares-data-closed`

data/emotes.json
SohKey: `data/emotes.json` is the same file that is in the resource pack, the `"providers"` is required by minecraft.

src/player/chat.js
SohKey: I took it off because it's ugly in the game, Delta also told me it was important but not why. I now that `' >> '` is not looking good but i dont know how to do it better

data/emotes.json
Sceat: hum then this should be in the resourcepack repo ? let's discuss this on discord with @DeltaEvo and create an issue about it as we may have 2 ressourcepack repo very soon

src/player/chat.js
DeltaEvo: The usage of the translation key is that this is something that we can control from the resource pack and this is the standard minecraft key for the chat format so it provides and extension point for resource packs.

If we want to change it for AresRPG we can do that by updating inside the resource pack lang file https://gist.github.com/DeltaEvo/fe3143497b6422b718dcb68d3744171a#file-en_us-json-L378

data/emotes.json
```diff
@@ -0,0 +1,25 @@
...
+    {
+      "type": "bitmap",
+      "file": "aresrpg:memes/emotes/ban.png",
+      "ascent": 9,
+      "height": 10,
+      "chars": [""]
```
DeltaEvo, line 8: We should create translation keys for the client like `aresrpg.emoji.ban` to not need to have that in this repo and only keep a list of emotes inside the `data` repository or just convert every :name: into a translation key, but I'm not sure what the client will do with that

src/player/chat.js
DeltaEvo: And further by re looking at wiki.vh, I remember with it's important, it's because this translation key is changed when the text is passed into the narrator which change the "style" formatting into text
```json
  "chat.type.text": "<%s> %s",
  "chat.type.text.narrate": "%s says %s",
```

> As a special case, if the translation key is chat.type.text, it will be changed to > chat.type.text.narrate when passed to narrator (although it will remain as
> chat.type.text in chat). 

https://wiki.vg/Chat#Translation_component

src/player/chat.js
```diff
@@ -73,6 +74,32 @@ export default {
       },
     }
 
+    const emoteFormated = emotes.providers.map(emote => {
```
DeltaEvo, line 77: nit: `emoteFormatted`

data/emotes.json
```diff
@@ -0,0 +1,25 @@
+{
```
DeltaEvo, line 1: Could you make a PR to the resource pack repository with this file ?

SohKey:
Bug: bad font when a none existing emote

Sceat (approved):

Sceat:
please rebase your branch

SohKey:
we have to add the emote json into the data submodules

SohKey:
Also this PR require merging:
- #aresrpg/data#3
- #aresrpg/ressourcepacks#7

src/player/chat.js
```diff
@@ -71,11 +94,10 @@ export default {
         }
         return chat
       },
+      [/:.*:/.source]: emote => emote_to_chat(emote.match(/.*/)[0]), // all emotes name are between ":"
```
DeltaEvo, line 97:
```suggestion
      [/:.*:/.source]: emote_to_chat, // all emotes name are between ":"
```

src/player/chat.js
```diff
@@ -34,6 +35,28 @@ function slot_to_chat({ nbtData, itemCount, itemId }) {
   return chat
 }
 
+function emote_to_chat(emote) {
+  const emote_name = `${emote.split(':')[1]}`
```
DeltaEvo, line 39:
```suggestion
  const emote_name = emote.slice(1, -1)
```

DeltaEvo (changes requested):

Sceat:
lol nice try hard 
image

but the 4th commit is still not [conventional](https://www.conventionalcommits.org/en/v1.0.0/)

Sceat (changes requested):
https://github.com/aresrpg/aresrpg/pull/571#issuecomment-1372473817

Sceat (approved):

PR #572, opened 2022-12-05
Handle right click on a bell

Arkobalt:
My first PR ever, don't really know how to structure it.
The feature is quite simple so i dont really need to explain it, i guess.

### DIFF
--- a/src/context.js
+++ b/src/context.js
@@ -27,6 +27,7 @@ import player_respawn from './player/respawn.js'
 import player_bossbar from './player/boss_bar.js'
 import player_action_bar from './player/action_bar.js'
 import player_heartbeat from './player/heartbeat.js'
+import player_bells from './player/bells.js'
 import player_traders, {
   register as register_player_traders,
 } from './player/traders.js'
@@ -266,6 +267,7 @@ export async function observe_client(context) {
   player_bossbar.observe(context)
   player_respawn.observe(context)
   player_heartbeat.observe(context)
+  player_bells.observe(context)
 
   commands_declare.observe(context)
 
--- a/src/player/bells.js
+++ b/src/player/bells.js
@@ -1,0 +1,29 @@
+import { get_block } from '../chunk.js'
+import { CATEGORY, play_sound } from '../sound.js'
+
+const BELL_ACTION_ID = 1
+const BELL_BLOCK_ID = 667
+
+export default {
+  /** @type {import('../context.js').Observer} */
+  observe(context) {
+    const { client, world } = context
+    client.on('block_place', async ({ location, direction }) => {
+      const block = await get_block(world, location)
+      if (block.name === 'bell' && direction >= 2) {
+        client.write('block_action', {
+          location,
+          byte1: BELL_ACTION_ID,
+          byte2: direction, // handle side from wich the bell is clicked (0 et 1 for up and down and other number for each side)
+          blockId: BELL_BLOCK_ID,
+        })
+        play_sound({
+          client,
+          sound: 'block.bell.use',
+          category: CATEGORY.AMBIENT,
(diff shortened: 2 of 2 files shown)


src/player/bells.js
```diff
@@ -0,0 +1,82 @@
...
+      if (block.name === 'bell') {
+        switch (
+          direction // bell face clicked
+        ) {
+          case 2: // north
+            client.write('block_action', {
+              location,
+              byte1: 1,
+              byte2: 2,
+              blockId: 677,
```
Sceat, line 19: those 3 values are magic numbers

src/context.js
```diff
@@ -267,6 +268,8 @@ export async function observe_client(context) {
   player_respawn.observe(context)
   player_heartbeat.observe(context)
 
+  player_bells.observe(context) // fezrhnjhifhbredjihebghfrvdihje
```
Sceat, line 271: `fezrhnjhifhbredjihebghfrvdihje` ?

Sceat (changes requested):
Thank for contributing to AresRPG !

About your PR, you have a huge code duplication, you could simply
```js
client.write('block_action', {
    location,
    byte1: 1,
    byte2: direction,
    blockId: 677,
  })
  play_sound({
    client,
    sound: 'block.bell.use',
    category: CATEGORY.AMBIENT,
    ...location,
  })
```

I guess creating a file just for the bells are fine @DeltaEvo ? but maybe it should be in a more generic world entity files ? like trapdoors etc (cc @KalvinVilla)

src/player/bells.js
Arkobalt: I didn't think about it x).
The slight difference with this version is that the bell ring (sound) even when there is no animation (when rang from top or bottom).
I son't know the exact details of what was required but maybe we'll need to add another condition for sound if done this way.

src/context.js
Arkobalt: It took me a day to realise that this was about the comment.
It's fixed.

src/player/bells.js
```diff
@@ -0,0 +1,28 @@
...
+          byte2: direction, // handle side from wich the bell is clicked (0 et 1 for up and down and other number for each side)
+          blockId: 677, // bell id
+        })
+        play_sound({
+          client,
+          sound: 'block.bell.use',
+          category: CATEGORY.AMBIENT,
+          x: location.x,
+          y: location.y,
+          z: location.z,
```
Sceat, line 23:
```suggestion
          ...location
```

Sceat (changes requested):

src/player/bells.js
```diff
@@ -0,0 +1,28 @@
...
+  observe(context) {
+    const { client, world } = context
+    client.on('block_place', async ({ location, direction }) => {
+      const block = await get_block(world, location)
+      if (block.name === 'bell' && direction >= 2) {
+        client.write('block_action', {
+          location,
+          byte1: 1, // only number 1 is available for bell animation
+          byte2: direction, // handle side from wich the bell is clicked (0 et 1 for up and down and other number for each side)
+          blockId: 677, // bell id
```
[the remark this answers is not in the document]
Arkobalt: oh, i see, i'll do that now

Sceat:
Seems good, let's now wait for @DeltaEvo opinion on https://github.com/aresrpg/aresrpg/pull/572#pullrequestreview-1205052568

Arkobalt:
My last commit was helpfull for test but may cause probleme because the mobs.json file from floor1 when starting the server. I dont know how without requesting a pull on floor1 but modifications on mobs.json should be discarded or the last commit should be ignored.

Sceat (approved):

PR #580, opened 2022-12-20
Improve contributing doc & notice M1 issue & bump deps

Sceat:


Sceat:
The PR is fine but weird CI errors that i don't get locally @DeltaEvo

PR #585, opened 2022-12-22
Add support for integration tests

Sceat:


test/player/resource_pack.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default function ({ doubt }) {
+  const bot = createBot({
+    host: 'localhost', // minecraft server ip
+    version: '1.16.5',
+    username: 'test',
+  })
+
+  return new Promise((resolve, reject) => {
+    bot.once('resourcePack', (url, hash) => {
```
Sceat, line 11: actually I should have used the promisified `once` here, will change

test/index.test.js
```diff
@@ -0,0 +1,31 @@
+import { pipeline, PassThrough } from 'stream'
+
+import { iter } from 'iterator-helper'
+import Doubt from '@hydre/doubt'
```
DeltaEvo, line 4: Could we use the standard node test runner instead ?
https://nodejs.org/api/test.html

test/player/resource_pack.test.js
```diff
@@ -0,0 +1,22 @@
...
+      doubt['should receive the correct resource pack']({
+        because: { url, hash },
+        is: {
+          url: 'https://github.com/aresrpg/resourcepacks/releases/download/v1.0.1/addon.zip',
```
DeltaEvo, line 15: Hardcoding that here is very brittle, we should only check that we got a ressource pack and at best that it's downloadable

test/player/resource_pack.test.js
```diff
@@ -0,0 +1,22 @@
+import { createBot } from 'mineflayer'
+
+export default function ({ doubt }) {
+  const bot = createBot({
+    host: 'localhost', // minecraft server ip
+    version: '1.16.5',
```
DeltaEvo, line 6: Should we not use the version from the settings ?

test/index.test.js
```diff
@@ -0,0 +1,31 @@
...
+const doubt = Doubt({
+  stdout: through,
+  title: 'AresRPG integration',
+  calls: 1,
+})
+
+process.env.LOG_LEVEL = 'silent'
+
+// main aresrpg server
+const { default: server } = await import('../src/index.js')
```
DeltaEvo, line 22: Should we recreate the server for each test ?

test/index.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+pipeline(through, reporter(), process.stdout, () => {})
+
+const doubt = Doubt({
+  stdout: through,
+  title: 'AresRPG integration',
+  calls: 1,
+})
+
+process.env.LOG_LEVEL = 'silent'
```
DeltaEvo, line 19: Not sure that's what we want here

test/index.test.js
```diff
@@ -0,0 +1,31 @@
...
+const { default: server } = await import('../src/index.js')
+// fastify debug webserver for mob behaviors
+const { debug } = await import('../src/context.js')
+
+await iter([resource_pack_suite])
+  .toAsyncIterator()
+  .forEach(suite => suite({ doubt }))
+
+server.close()
+debug.app.close()
```
DeltaEvo, line 31: We should maybe listen for closes of the server inside debug.js instead

DeltaEvo (changes requested):

test/index.test.js
Sceat: With "doubt" this one isn't a suite but the entrypoint so it only run once. But I'm gonna check the node test runner

test/index.test.js
Sceat: We kinda don't want to whole server log conflicting with tests logs, it's there if someone have failing tests and need to dig but in anycase the output is not formatted

test/index.test.js
Sceat: good catch

src/behavior.js
```diff
@@ -51,8 +50,11 @@ export default async function run(node, state, context) {
     path,
   }
   const result = await nodes[node.tagName](node, state, node_context)
-  log.debug({ path, status: result.status.toString() }, 'Runned')
-  debug.behavior({ context: node_context, result })
+  log.debug({ path, status: result.status.toString() }, 'Ran')
+
+  const debug_server = context.world.get_debug()
+  if (debug_server) debug_server.behavior({ context: node_context, result })
```
DeltaEvo, line 56: I would prefer that we keep this global debug and that we make it null when a config option is set instead of hacking it inside the world

src/context.js
```diff
@@ -81,7 +80,6 @@ const initial_world = {
   next_entity_id: 1,
   next_window_id: 1, // 0 is the player inventory
   /** @type {() => Object} Remove type to remove circular references */
-  get: () => world,
```
DeltaEvo, line 84: You need to keep that here, otherwise the infered typescript type will not be the same

src/context.js
```diff
@@ -96,16 +94,8 @@ const world_reducers = [
   register_player_item_loot,
 ]
 
-const world = /** @type {World} */ (
```
DeltaEvo, line 99: The world by itself is immutable, you want to keep that and instead just recreate a new event emitter if you want to create multiple instances of it

src/context.js
```diff
@@ -96,16 +94,8 @@ const world_reducers = [
...
-  world_reducers.reduce(
-    (world, fn) => fn(world),
-    /** @type {any} */ (initial_world)
-  )
-)
-
 const initial_state = {
   nickname: undefined,
-  position: world.spawn_position,
```
DeltaEvo, line 108: This can stay here or be transformed into floor1.spawn_position

src/context.js
```diff
@@ -223,164 +213,179 @@ function transform_action(action) {
   ].reduce((intermediate, fn) => fn(intermediate), action)
 }
 
-const mobs_position = mobs_position_factory(world)
-
-/** @type Observer */
-export async function observe_client(context) {
```
DeltaEvo, line 229: Could we not pass all the dependencies as arguments of this function instead ? The world and the mobs_position observer and create thoses in the server.js ?

src/debug.js
```diff
@@ -16,14 +16,14 @@ const log = logger(import.meta)
 /**
  * @param {{ world: any, app: import('fastify').FastifyInstance }} param
  */
-function behavior({ world, app }) {
+function behavior({ mobs, app }) {
```
DeltaEvo, line 19: I would prefer us to keep the world here as we can extend it to read other properties

src/player/item_loot.js
```diff
@@ -29,28 +31,33 @@ export function register(world) {
...
-  client.write('entity_metadata', {
-    entityId: entity_id,
-    metadata: to_metadata('item_entity', {
-      entity_flags: { has_glowing_effect: true },
-      item: item_to_slot(items[item.type], item.count),
-    }),
-  })
+  const aresrpg_item = items[item.type]
+
+  if (aresrpg_item) {
```
DeltaEvo, line 36: We should not have invalid items in the first place, please remove that

test/player/resource_pack.test.js
```diff
@@ -0,0 +1,101 @@
+import assert from 'assert'
+import { once } from 'events'
+import { after, before, describe, it } from 'node:test'
+
+import { createBot } from 'mineflayer'
+
+import { create_server } from '../../src/server.js'
+import { VERSION } from '../../src/settings.js'
+
+const KICK_REASON = [
```
DeltaEvo, line 10: We should not repeat this here

test/player/resource_pack.test.js
```diff
@@ -0,0 +1,101 @@
...
+describe('The resource pack', () => {
+  // mandatory mutations
+  const options = {
+    server: null,
+    // this function comes from the create_server()
+    client_by_username: null,
+    find_user: null,
+  }
+
+  before(() => {
```
DeltaEvo, line 47: This should be a beforeEach, afterEach.

I would even prefer for us to create the server in the test and pass it as argument a signal https://nodejs.org/api/test.html#contextsignal to close rhe server when the test is finished

test/player/resource_pack.test.js
```diff
@@ -0,0 +1,101 @@
...
+  })
+
+  it('should be a valid link', async () => {
+    const bot = createBot({
+      host: 'localhost',
+      version: VERSION,
+      username: 'aboubacar',
+    })
+    const [url] = await once(bot, 'resourcePack')
+    const { ok } = await fetch(url, { method: 'HEAD' })
```
DeltaEvo, line 62: Should we mock the ressource pack url to not depend on the network ?
Maybe add a todo ?

test/player/resource_pack.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+    assert.ok(ok)
+
+    bot.end()
```
DeltaEvo, line 66: nit: use a signal in createBot instead https://nodejs.org/api/test.html#contextsignal

test/player/resource_pack.test.js
```diff
@@ -0,0 +1,101 @@
...
+      // seems that accents are not supported by the bot
+      // so we replace them by simple letters
+      JSON.parse(reason.normalize('NFD').replace(/[\u0300-\u036f]/g, ''))
+    )
+
+    await once(bot, 'resourcePack')
+
+    // this part was a bit tricky
+    // we have to access the client object of the bot
+    const client = options.client_by_username(username)
```
DeltaEvo, line 87: This seems weird and deems more to be a bug in node minecraft protocol, could you describe more what is happening ?

DeltaEvo (changes requested):

src/player/item_loot.js
Sceat: this means removing all items and mobs from the public world as they don't exist anymore, guess I'll do that

test/player/resource_pack.test.js
Sceat: hum this is arguable, it is the expected kick reason so if I source it from the code the test isn't that meaningful

test/player/resource_pack.test.js
Sceat: I think it's fine during integration tests to depend on the network, it's even a good sign to catch problems early. Only in unit tests we shall be noop

src/player/item_loot.js
DeltaEvo: Well we don't care about the actual reason, only that the player was kicked if it refused the pack.
Asserting the reason make the test very britle making it fail when we modify the message where it should not as this will still function even if we modify the message.

test/player/resource_pack.test.js
DeltaEvo: Well we don't care about the actual reason, only that the player was kicked if it refused the pack.
Asserting the reason make the test very britle making it fail when we modify the message where it should not as this will still function even if we modify the message.

test/player/resource_pack.test.js
DeltaEvo: The question then is this a unit or an integration test ? Because the boundary is very thin.
What do we consider as a unit ?

But it's not the subject, I'm mostly nitpicking, fine for me, you can resolve this comment

test/player/resource_pack.test.js
Sceat: depends if we test the concept of being kicked or the full intended kick including a strict message, both are fine so let's remove it to have a smaller code

test/player/resource_pack.test.js
Sceat: unit tests are directly testing some functions, while integration include a wider behavior. It's nice to do both in Ares and it will just be differentiated as

## Unit
import the function from `src`, no IO

## Integration
run the server and only test the bot behavior

test/player/resource_pack.test.js
Sceat: createBot doesn't take a signal in its options

test/player/resource_pack.test.js
Sceat: actually the signal is never sent so we still need the afterEach to close the server

src/server.js
```diff
@@ -0,0 +1,138 @@
...
+  register_mobs,
+  register_mobs_position,
+  register_player_traders,
+  register_player_deal_damage,
+  register_experience,
+  register_player_teleportation_stones,
+  register_player_item_loot,
+]
+
+export const world = /** @type {World} */ (
```
DeltaEvo, line 59: Keep the world inside context.js

src/context.js
```diff
@@ -223,78 +177,77 @@ function transform_action(action) {
...
-  mobs_look_at.observe(context)
-  mobs_wakeup.observe(context)
-  mobs_loot.observe(context)
-  mobs_attack.observe(context)
-  mobs_sound.observe(context)
-
-  chunk_update.observe(context)
+export function observe_client({ mobs_position }) {
+  /** @type Observer */
+  return async context => {
```
DeltaEvo, line 182: Take two arguments, when don't need to curry here

src/context.js
```diff
@@ -352,6 +310,7 @@ export async function create_context(client) {
       // default nickname is the client username, and is overriden by the loaded player state
       {
         ...initial_state,
+        position: world.spawn_position,
```
DeltaEvo, line 313: Remove

src/debug.js
```diff
@@ -90,14 +92,14 @@ function behavior({ world, app }) {
...
-export default function start_debug_server({ world }) {
+function start_debug_server() {
   const app = fastify({ logger: log })
 
   app.register(cors, {
     origin: true,
   })
 
-  app.listen({ port: 4242 }).then(address => {
+  app.listen().then(address => {
```
DeltaEvo, line 102: Keep the port as is, it plays better with your browser history, if we have port clashes we could do something in the future to try to find a port near to the one we are used to (4243, ...)

src/mobs/position.js
```diff
@@ -54,7 +54,7 @@ export function register(world) {
   }
 }
 
-export default function observe_world({ mobs }) {
+export default function mobs_position_factory({ mobs }) {
```
DeltaEvo, line 57: Why did you renamed that ?

src/server.js
```diff
@@ -0,0 +1,138 @@
...
+  world_reducers.reduce(
+    (world, fn) => fn(world),
+    /** @type {any} */ (initial_world)
+  )
+)
+
+export const create_server = ({ signal } = {}) => {
```
DeltaEvo, line 66: Use a function instead of an arrow function

DeltaEvo (changes requested):

src/mobs/position.js
Sceat: to keep consistency with import names. I like using default export but when the file is not a function like
```
get_x.js
```
```
import get_x from 'get_x.js'
```
the minimum required should be at least to name it the same way we use it, and here you used `mobs_position_factory` in the context. Otherwise we should not use default export as it will be confusing

src/context.js
Sceat: it allows me to
```js
create_context({ client, world })
      .then(observe_client({ mobs_position }))
```

Sceat:
PR is fully done, except for those broken typings
image

DeltaEvo (approved):

PR #601, opened 2023-01-04
Basic class selection menu

loucass003:
javaw_hY5zhWeKKi

We still need to define what items to give the player based on the selection and do proper textures for the kit but everything is there

### DIFF
--- a/src/context.js
+++ b/src/context.js
@@ -44,6 +44,7 @@ import player_item_loot, {
   register as register_player_item_loot,
   ITEM_LOOT_MAX_COUNT,
 } from './player/item_loot.js'
+import player_class_selection from './player/class_selection.js'
 import player_soul from './player/soul.js'
 import finalization from './finalization.js'
 import plugin_channels from './plugin_channels.js'
@@ -117,6 +118,8 @@ const initial_state = {
   inventory_cursor_index: 0,
   held_slot_index: 0,
   game_mode: GameMode.ADVENTURE,
+  class_selection_open: false,
+  selected_class: 0,
   experience: 0,
   health: 40,
   // player's energy, losing after each death
@@ -140,6 +143,7 @@ const saved_state = ({
   health,
   soul,
   last_disconnection_time,
+  selected_class,
 }) => ({
   nickname,
   position,
@@ -150,6 +154,7 @@ const saved_state = ({
   health,
   soul,
   last_disconnection_time,
+  selected_class,
 })
 
 /** @template U
@@ -188,6 +193,7 @@ function reduce_state(state, action) {
     player_soul.reduce,
     player_health.reduce,
     player_experience.reduce,
+    player_class_selection.reduce,
     chunk_update.reduce,
   ].reduce((intermediate, fn) => fn(intermediate, action), state)
 }
@@ -235,6 +241,7 @@ export function observe_client({ mobs_position }) {
     player_respawn.observe(context)
     player_heartbeat.observe(context)
     player_bells.observe(context)
+    player_class_selection.observe(context)
 
     commands_declare.observe(context)
 
--- a/src/events.js
+++ b/src/events.js
@@ -24,6 +24,10 @@ export const PlayerAction = {
   SWITCH_GAMEMODE: 'PLAYER:ACTION:SWITCH_GAMEMODE',
   /** the player should receive experience */
   RECEIVE_EXPERIENCE: 'PLAYER:ACTION:RECEIVE_EXPERIENCE',
+  /** open the class selection menu */
+  OPEN_CLASS_SELECTION: 'PLAYER:ACTION:OPEN_CLASS_SELECTION',
+  /** open the class selection menu */
+  SELECT_CLASS: 'PLAYER:ACTION:SELECT_CLASS',
 }
 
 export const PlayerEvent = {
--- a/src/player/class_selection.js
+++ b/src/player/class_selection.js
@@ -1,0 +1,167 @@
+import { on } from 'events'
+import { setInterval } from 'timers/promises'
+
+import combineAsyncIterators from 'combine-async-iterators'
+import { aiter } from 'iterator-helper'
+
+import { PlayerAction, PlayerEvent } from '../events.js'
+import { abortable } from '../iterator.js'
+import { delete_title, write_action_bar, write_title } from '../title.js'
+import { GameMode } from '../gamemode.js'
+
+const KITS_AMOUNT = 4
+
+function send_selection_menu(client, selected) {
+  const kits = Array.from({ length: KITS_AMOUNT }).map((_, i) => ({
+    translate: `aresrpg.kits.kit.${i}`,
+    font: 'kits:default',
+  }))
+
+  kits.splice(selected + 1, 0, {
+    translate: 'space.-65',
+    font: 'space:default',
+    // @ts-ignore
+    with: [
+      {
+        translate: 'aresrpg.kits.selected',
+        font: 'kits:default',
+      },
+    ],
+  })
+  write_title({
+    client,
+    subtitle: kits,
+    times: {
+      fade_in: 0,
+      fade_out: 0,
+      stay: 5,
+    },
+    force_reset: false,
+  })
+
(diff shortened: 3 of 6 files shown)


loucass003:
Also need https://github.com/aresrpg/resourcepacks/pull/9 to be used/tested

src/player/class_selection.js
```diff
@@ -0,0 +1,167 @@
...
+          state,
+          {
+            class_selection_open,
+            held_slot_index,
+            timer,
+            game_mode,
+            selected_class,
+          }
+        ) => {
+          const { last_held_slot_index, last_open, last_selected_class } = state
```
Sceat, line 103: A bit ambiguous to call the reduced elements `state`, would be less confusing to extract everything directly

src/player/class_selection.js
```diff
@@ -0,0 +1,167 @@
...
+
+          const new_selected_class = slot_next_value(
+            held_slot_index,
+            last_held_slot_index,
+            selected_class
+          )
+
+          if (
+            last_held_slot_index !== held_slot_index &&
+            class_selection_open
```
Sceat, line 118: probably this should be filtered out so that this reducer don't trigger when there is no class selection

src/player/class_selection.js
```diff
@@ -0,0 +1,167 @@
...
+        }
+      )
+
+    client.on('arm_animation', ({ hand }) => {
+      events.once(
+        PlayerEvent.STATE_UPDATED,
+        ({ selected_class, class_selection_open }) => {
+          if (class_selection_open && hand === 0) {
+            dispatch(PlayerAction.OPEN_CLASS_SELECTION, { open: false })
+            // TODO: give player the selected class
```
Sceat, line 161: instead of a comment this should just dispatch the `SELECT_CLASS`

src/events.js
```diff
@@ -24,6 +24,10 @@ export const PlayerAction = {
   SWITCH_GAMEMODE: 'PLAYER:ACTION:SWITCH_GAMEMODE',
   /** the player should receive experience */
   RECEIVE_EXPERIENCE: 'PLAYER:ACTION:RECEIVE_EXPERIENCE',
+  /** open the class selection menu */
+  OPEN_CLASS_SELECTION: 'PLAYER:ACTION:OPEN_CLASS_SELECTION',
+  /** open the class selection menu */
```
Sceat, line 29: wrong comment

src/player/class_selection.js
```diff
@@ -0,0 +1,167 @@
...
+          )
+
+          if (
+            last_held_slot_index !== held_slot_index &&
+            class_selection_open
+          ) {
+            send_selection_menu(client, new_selected_class)
+            dispatch(PlayerAction.SELECT_CLASS, {
+              selected_class: new_selected_class,
+            })
```
Sceat, line 123: You can store the currently selected class after the player choose it, but i don't think you need to store the currently scrolled one as you can use the reducer for that

src/title.js
```diff
@@ -22,17 +22,20 @@ export function write_action_bar({ client, text }) {
...
  * Send a title to a client, all times are in seconds
- * @param {{ client: any, title?: any, subtitle?: any, times?: any }} arg
+ * @param {{ client: any, title?: any, subtitle?: any, times?: any, force_reset?: boolean }} arg
  */
 export function write_title({
   client,
   title = empty_component,
   subtitle = empty_component,
   times,
+  force_reset = true,
```
Sceat, line 32: Seems cleaner to actually export another function which just clears the title and use it wherever this one was legacy used

Sceat (changes requested):

src/player/class_selection.js
loucass003: The reason io do not extract the reduce state directly is because of
```js
if (timer) {
  if (last_open) send_selection_menu(client, last_selected_class)
  return state
}
```

in this case because i am not mutating state it is easier to return it directly than reconstructing it from all the fields that i extracted

src/player/class_selection.js
loucass003: Not sure what you mean. it is already filtered, that is what this condition is for. I cannot filter it on a higher level than this because this reducer also handle the visibility of the slection menu. So i need an event when the menu both open and close

src/player/class_selection.js
loucass003: problem is that i have no way to get the curently selected class from the arm_animation event without storing it inside the player state

src/player/class_selection.js
loucass003: I cant access the curently selected class from here. so this is why i made it like that

src/context.js
```diff
@@ -117,6 +118,8 @@ const initial_state = {
   inventory_cursor_index: 0,
   held_slot_index: 0,
   game_mode: GameMode.ADVENTURE,
+  class_selection_open: false,
+  selected_class: 0,
```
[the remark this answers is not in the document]
loucass003: i would rather keep it as an index. because we would get those values from a json file anway. The code deal with the selection as index and not a string or object. Until we know exactly was a class is going to be i dont think we need more than this.

src/player/login.js
```diff
@@ -91,6 +91,9 @@ export default {
             { text: ' for the first time', ...Formats.BASE },
           ],
         })
+
+        dispatch(PlayerAction.OPEN_CLASS_SELECTION, { open: true })
```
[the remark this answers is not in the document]
loucass003: Sure. This was writen anywhere that we could select multiple classes. So i didnt do it like that. But again we do not have discussed about how any of this is stored. This code can be changed after once we actually decided what to do.
The most important part of this is that we can now do negative caracters menus on the screen. and are able to update it when needed. Having other hover/selected state on top of it will be easy to add.

Unless you can provide me a DETAILED issue with how everything should work, step by step. and how we are planing to store those value. i refuse to go any further on this PR because going back and forth would be too unproductive

src/title.js
loucass003: Just verified. your are right. on top of that i just checked and there is not places rn that requires the use of the reset title action anyway. i added a function just i case wee need it. but we can safely remove it here

src/context.js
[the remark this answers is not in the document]
Sceat: we actually already know https://github.com/aresrpg/aresrpg/discussions/categories/classes

src/player/class_selection.js
Sceat: alright let's then use something like `payload` or `result`, to avoid reusing `state` and `context` which we heavily use for other purposes in the codebase

src/player/class_selection.js
Sceat: unless we rely on specific slots index instead of infinite scrolling, we have only 4 classes and the player will have a maximum of 6 characters, so the 9 hotbars slots are enough. It would also allow selecting class/characters with the numkeys (playing with trackpad)

src/player/class_selection.js
Sceat: https://github.com/aresrpg/aresrpg/pull/601/files/b5a92b5928554dea1d19d77d40d17f05b22a12f2#r1070833129

src/player/login.js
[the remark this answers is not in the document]
Sceat: 🫡 https://github.com/aresrpg/aresrpg/issues/604

src/player/class_selection.js
```diff
@@ -0,0 +1,166 @@
...
+  }))
+
+  kits.splice(selected + 1, 0, {
+    translate: 'space.-65',
+    font: 'space:default',
+    // @ts-ignore
+    with: [
+      {
+        translate: 'aresrpg.kits.selected',
+        font: 'kits:default',
```
SohKey, line 27:
```suggestion
    font: 'aresrpg:kits',
  }))

  kits.splice(selected + 1, 0, {
    translate: 'space.-65',
    font: 'space:default',
    // @ts-ignore
    with: [
      {
        translate: 'aresrpg.kits.selected',
        font: 'aresrpg:kits',
```

Issue #444, opened 2022-01-23
Some trapdors should be openable


DeltaEvo:
Implement openable trapdors

Arkobalt:
Aren't trapdoors already openable ?

Sceat:
It's client side only, and we may want to only allow certain trap doors and not all ? @DeltaEvo

Sceat:
> Aren't trapdoors already openable ?

Probably client side trapdoors are enough, this issue isn't much detailed so let's close it for now

PR #602, opened 2023-01-04
Interactable object 

KalvinVilla:
This PR respond at issue `some trapdors should be openable`
but with more interactible object (door, fence ...)


Sceat (changes requested):
Cleaning the history doesn't mean deleting the branch. It means you have to follow https://www.conventionalcommits.org/en/v1.0.0/ and have a clear sequence of meaningful addition with their proper names. As for this PR `Cleaning history :)` is not a meaningful commit

Sceat:
Actually for now #444 vanilla behavior is enough, let's not have this PR

Issue #396, opened 2021-12-02
Joining with an invalid version result with a disconnection after a timeout of 30 seconds


Sceat:
Settings to `1.16.3`
i join with `1.16.5`
i punch a dumb chicken
i wait
i get kicked

DeltaEvo:
It's really weird since https://github.com/PrismarineJS/minecraft-data/blob/master/data/dataPaths.json#L910 it should be the same protocol

DeltaEvo:
Ok so after investigating this is a bug in node-minecraft-protocol, the client is ended since the protocol version don't match https://github.com/PrismarineJS/node-minecraft-protocol/blob/1796b9693d02be5fa8e6e4a03c547e430b0bf21d/src/server/handshake.js#L18 and since we are changing the serializer the connection is closed by the closeTimer https://github.com/PrismarineJS/node-minecraft-protocol/blob/1796b9693d02be5fa8e6e4a03c547e430b0bf21d/src/client.js#L182

PR #605, opened 2023-01-17
fix: Joining with an invalid version

SohKey:
fix: #396

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -60,6 +60,24 @@ export function create_server() {
   const mobs_position = mobs_position_factory(world)
 
   server.on('login', client => {
+    const {
+      // @ts-ignore
+      mcversion: { minecraftVersion, version },
+    } = server
+
+    // check protocol versions
+    if (client.protocolVersion !== version) {
+      client.end(`Wrong minecraft version, expected: ${minecraftVersion}`)
+      log.info(
+        {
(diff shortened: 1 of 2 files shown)


src/server.js
```diff
@@ -59,6 +59,24 @@ export function create_server() {
   const mobs_position = mobs_position_factory(world)
 
   server.on('login', client => {
+    const {
+      // @ts-ignore
+      mcversion: { minecraftVersion, version },
```
SohKey, line 64: I dont know why `mcversion` is not recognized, i'm forced to use ts-ignore

Sceat (changes requested):
This issue is more about fixing the problem in prismarine or wherever it is, and allowing players to join regardless of their minor versions

Sceat:
@SohKey I guess you can make some research work in here, but if it's not fixable on AresRPG, this pr should be closed

Sceat:
After some thinking, this is actually okay to have as a temporary fix. @SohKey can you just improve the `end` message similarly to https://github.com/aresrpg/aresrpg/blob/master/src/player/resource_pack.js

Note that later on we'll need to find a way to support multiple version of the same protocol

Sceat (approved):
lgtm

Sceat (approved):

Issue #302, opened 2021-09-02
Implement Immersive sounds


Sceat:
with datapack i guess, let's see for the in game music if we use noteblocks

immersive datapack: https://www.curseforge.com/minecraft/customization/immersive-sounding

sounds overhaul: https://youtu.be/UMq8zHyUN8k

DeltaEvo:
A datapack is server side, so in the ressourcepack instead :)
And I don't think using noteblock sounds is a good idea

Sceat:
noteblocks might be the only way to not force 10go of download to the client

DeltaEvo:
Noteblocks song will be much heavier that a resource pack song, a song in a resource pack is encoded using a specialised structure, with noteblock each note is wrapped in a packet

Sceat:
On the big v1 pack there was music for all cities it seems https://www.youtube.com/watch?v=LDiqnvv_8e8 😭

ToufouMaster:
maybe creating a special resource pack for that.
When you login you get the pack and if in you activate the last time you played the /immersive command the you will download the immersive pack every time you login

Sceat:
So you're saying we can enforce multiple server pack ? not sure about that

ToufouMaster:
why not? and if not, remove the first one and erplace it with the new that contain immersive sounds

SohKey:
this can help: [minecraft sounds keys](https://pokechu22.github.io/Burger/1.19.html#sounds)

SohKey:
we can't add new songs without replacing vanilla's ones, so we have to replace blocks sound (or anyone that a player can't trigger)

SohKey:
exemple for sounds.json doc [here](https://minecraft.fandom.com/wiki/Sounds.json)
```
"example.sound.name": {
    "category": "hostile",
    "sounds": [
      {
        "name": "whatever/path/example_3",
        "volume": 0.5,
        "pitch": 2.5,
        "stream": true,
        "preload" true
      },
      {
        "name": "whatever/path/example_4",
        "volume": 0.2,
        "pitch": 0.8,
        "attenuation_distance": 15
      },
      {
        "name": "whatever/path/example_5",
        "volume": 1.5,
        "pitch": 1.5,
        "weight": 3
      }
    ]
  }
```

Issue #37, opened 2020-11-01
Study custom biomes and dimensions for aresrpg


DeltaEvo:
1.16 introduced custom dimensions and 1.16.2 custom biomes (thanks @Vahelnir)

Maybe we could use this for AresRPG

On the protocol it's implemented as a nbt dimensionCodec https://github.com/aresrpg/aresrpg/blob/master/src/world/codec.js and sent in the login packet https://github.com/aresrpg/aresrpg/blob/master/src/login.js#L23

Some initial links:
https://www.minecraft.net/en-us/article/minecraft-java-edition-1-16-2
https://www.planetminecraft.com/data-pack/custom-biome-datapack-20w28a-higher/

DeltaEvo:
https://github.com/slicedlime/examples/blob/master/vanilla_worldgen.zip

DeltaEvo:
https://minecraft.gamepedia.com/Custom_dimension

Sceat:
@SohKey for the zones feature

SohKey:
> @SohKey for the zones feature

thx

Sceat:
@SohKey we need to have a `zone.json` inside the World repo, that could be defined either manually or generated by a tool (ingame, mcedit, etc..), those zone will be used for:

- display zone information
- have a zaap somewhere inside (tp stone)
- have specific entities spawn datas
- prevent mob from switching zones (working in pair with #645) cc @DeltaEvo 
- specific music
- specific ambiance/sky color

PR #607, opened 2023-01-18
Add immersive sounds

SohKey:
#302


SohKey:
Need merging  

> aresrpg/ressourcepacks#11

Sceat (changes requested):
It's not that simple, we should discuss on the issue for more details but the immersive sounds are about trigerring wind in valeys, bird and leaves in forests, waves on shores, and musics on cities/donjons. Which means a zone system need to be in place first along with their respective TP stones (which should become portal)

SohKey:
> It's not that simple, we should discuss on the issue for more details but the immersive sounds are about trigerring wind in valeys, bird and leaves in forests, waves on shores, and musics on cities/donjons. Which means a zone system need to be in place first along with their respective TP stones (which should become portal)

Okay i can do it, just need all the immersive sounds that you want, for the zone system i can use biomes and the actual tp stones as you sayed

SohKey:
> > It's not that simple, we should discuss on the issue for more details but the immersive sounds are about trigerring wind in valeys, bird and leaves in forests, waves on shores, and musics on cities/donjons. Which means a zone system need to be in place first along with their respective TP stones (which should become portal)
> 
> Okay i can do it, just need all the immersive sounds that you want, for the zone system i can use biomes and the actual tp stones as you sayed

i you don't have the sounds i can find them

Sceat:
We don't have the sounds so I guess we should simply have a zone system with an optional sound/music attached to it, I'm not sure how the biomes works in detail and if we can use that (#37), a forest could be a zone so it'll have its specific sound etc

src/player/resource_pack.js
```diff
@@ -23,6 +24,7 @@ export default {
       const status = Object.entries(Status).find(
         ([key, value]) => value === result
       )
+      events.emit(PlayerEvent.PACK_LOAD, result)
```
Sceat, line 27: This seems wrong, we don't want to trigger music when the pack is sent, we could wait for the client to accept and fully load it but still, it would be a security and not a real feature.

src/player/sounds.js
```diff
@@ -0,0 +1,42 @@
...
+  /** @type {import('../context.js').Observer} */
+  observe({ client, events, world, signal }) {
+    events.once(PlayerEvent.PACK_LOAD, () => {
+      aiter(
+        abortable(on(events, PlayerEvent.STATE_UPDATED, { signal }))
+      ).reduce(
+        ({ last_played_song }, [state]) => {
+          const { position } = state
+          const music =
+            closest_stone(world, position)?.music ?? 'aresrpg.ambient.none'
```
Sceat, line 21: It's not about absolutely trying to play some music, the feature should be plugged on top of a real well made zone feature and music would sometimes play but not always (like vanilla behavior)

Sceat (changes requested):
I understand the haste of "cleaning up smalls issues" but the current behavior wouldn't help us

src/player/sounds.js
SohKey: Okay is not a problem, i made it with tp stone for the towns, but for the donjons & forest ambiances, will add a new json.
And for the random trigger it's easy to do, but if you want me to know you want, please be more descriptive in the issue...

src/player/resource_pack.js
Sceat: It's alright to fail a music, it should sometimes appear and some other time disapear

SohKey: Okay but for the players it's like a bug, the music start and 1s after it's stopped. Are you sure that we want to delete that ?

Sceat: actually by looking at the file, this snippet is obsolete, it should be deleted as there already is the result handler below
image

doing `events.once(PlayerEvent.PACK_LOAD` in sounds.js is useless because when allowing the ressourcepack, the event is already awaited

```js
    if (USE_RESSOURCE_PACK) await player_resource_pack.observe(context)
```

SohKey: Im blind 😑

PR #608, opened 2023-01-18
Provide a data interface to allow optional submodules

Sceat:
It's getting really annoying to have multiple duplicated repo for public and private datas, let's reduce that to 1 and provide a fallback


Sceat:
obsolete

Issue #152, opened 2021-02-26
Soul system


Sceat:
## Update (2023)

Reopened to handle the complete feature.

- Each death remove 10%
- When reaching 0 of soul the player should become a ghost and walk very slowly, he shouldn't be able to interact with anything
- While being a ghost, the player should be invisible and have some sort of ghost model on top of him
- The player can turn back to life and with 10 of soul by reaching a phoenix statue 

<details>
<summary><h2>Original issue</h2></summary>

The soul on aresrpg is a way to penalize death.
![image](https://user-images.githubusercontent.com/11330271/109318673-5859e280-784e-11eb-8220-6440616fe520.png)

- Each death remove 10%
- Below 50% the player has a 50% experience gains penality
- At 0%, the player has a 90% experience penality and his stats are reduced by 40%
- The sould can be regenerated by paying an NPC or by waiting enough time
</details>

ToufouMaster:
this should be reworked a bit i think

Sceat:
we can implement a fantom state 😁 the only effect of the soul would be that at 0 the player become a fantom and be extremely slow until he reach a resurrection statue

ToufouMaster:
this is a good idea but i think it can be more upgraded
(losing something but What?)

Sceat:
That's infamous to burn player's stuff when they die, way more frustrating then reduced xp

Sceat:
#582 implements this except:
- the "revive" part as we need to be able to read the builder's map to access phoenix statues.
- The ghost model as we need our custom modelengine

PR #669, opened 2023-04-20
Thomas/biome update

SohKey:
Command to update the chunk's biome


src/chunk.js
```diff
@@ -88,3 +88,8 @@ export function chunks(region_folder) {
     },
   }
 }
+
+export function save_chunk({ region_folder, x, z, chunk }) {
+  const provider = new AnvilWorld(region_folder)
+  return provider.save(x, z, chunk)
```
Sceat, line 94: We can't do that from the server, the world should be readonly. I also need to modify the map for the #152 feature but not sure how for now, anyway even for the zone if it's through the world's file then we should do it externally

src/commands/commands.js
```diff
@@ -108,6 +109,9 @@ export default function execute_command({
     case 'soul':
       soul(command)
       break
+    case 'setBiome':
+      setBiome(command)
+      break
```
Sceat, line 114: yeah that's the thing AresRPG can't be an editor, we have to do those writes externally

src/commands/setBiome.js
```diff
@@ -0,0 +1,128 @@
...
+
+  const chunk = await world.chunks.load(x, z)
+  const biomes = chunk.dumpBiomes()
+
+  // update the biome foreach block in the chunk
+  biomes.forEach((value, index, array) => {
+    array[index] = biomeId
+  })
+
+  client.write('map_chunk', {
```
Sceat, line 48: won't this conflict with the chunk generation ? if we save the zones outside the `floor1` module then yea we can interfere with the chunk generation directly by loading updates from like a json file but it should be integrated in the server's chunk system

src/world/customBiomes.json
```diff
@@ -0,0 +1,166 @@
+{
```
Sceat, line 1: I suppose this is a generated file, as it's tied to the map i suppose it should be in the world module

Sceat (changes requested):
Let's follow up on discord for how we handle the source of new zones

src/chunk.js
SohKey: we need to save the modified chunk because when server restart it will remain unchanged. Maybe we can desactivate the command when we will finish to setup zones.

src/world/customBiomes.json
SohKey: yep

Issue #305, opened 2021-09-05
State is not saved when client error


Sceat:
Quite often when a player disconnect, an client error is thrown which prevent the state from being saved here
```js
  aiter(
    combineAsyncIterators(
      actions[Symbol.asyncIterator](),
      packets,
      ...blockchain.sources
    )
  )
    .map(transform_action)
    .reduce(
      (last_state, action) => {
        const state = reduce_state(last_state, action)
        events.emit('state', state)
        return state
      },
      { ...initial_state, ...player_state }
    )
    .then(save_state)
    .catch(error => {
      // TODO: what to do here if we can't save the client ?
      log.error(error, 'State error')
    })
```

how can we get the final state to call `save_state` in a `.finally` ?

DeltaEvo:
I observed that too. The root issue for me is that the "actions" stream and the reduce body should never throw errors, currently I guess that this is an unwanted behavior from `events.on` (it listen to the error event) on the packets stream

We could also as a mitigation (for future errors) for that to catch the error and try to save the get_state value, however I'm not a big fan of adding this now because we might in the future store the whole state updates of the player

PR #693, opened 2023-07-30
Prevent state errors on socket error

Sceat:
Closes: #305


Sceat:
There are still unhandled rejection happening, but I'm not able to pinpoint where :/

To trigger the error, you have to join the server then disconnect, the join again, until it crashes. 

image

PR #696, opened 2023-08-01
Support Modules compositions for AresRPG

Sceat:
📄 **Description**

A fundamental overhaul of our module structure aims to bolster simplicity and intuitive code navigation. This adjustment ensures a clear and purposeful direction of execution, transforming the way modules are implemented. This allows functionalities in Aresrpg to be toggled with ease.

🚀 **Motivation**

This enhancement is vital for issue #695. It provides a more versatile approach in handling various game states and scenarios a player might encounter:
- 🎮 Main menu & character selection
- 👻 Ghost mode
- 🏦 Inside the bank
- 🔧 Admin mode, etc.

Additionally, it streamlines the onboarding process for new developers.

🎬 **Starting Point**

We're currently revolving around 4 primary actors:
1. 🧟 The mobs
2. 🌍 The world
3. 🧑 The player
4. 🖥️ The server

> A module in AresRPG is a file that can manage a state and observe related events. Reducers ingest a state and a specific action, possibly adjusting the state based on the action. Observers wait for events and then dispatch actions which are managed by reducers. Initial reducers are executed once, setting up a state instance for runtime use.

🌲 **Structure**

```plaintext
AresRPG
│
├── 📑 index.js 
│   ├── Input: None
│   ├── Action: Connects all information sources
│   └── Output: Initialized server and modules
│
├── 🖥️ server.js 
│   ├── Input: Client listener 
│   ├── Action: Creates the server 
│   └── Output: Established server instance
│
├── 🌍 world.js 
│   ├── Input: Array of `initial_world_reducer`, `create_mob` function 
│   ├── Action: Crafts the world, maps mob positions 
│   └── Output: Augmented world instance
│
├── 🧟 mobs.js 
│   ├── Input: World, array of pathfinders 
│   ├── Action: Initializes mob states 
│   └── Output: Array of initialized mobs 
│
└── 🧑 player.js 
│   ├── Input: Array of player modules 
│   ├── Action: Sets up player state 
│   └── Output: Client handler function
│
├── 🔄 core/ - Cross-module static functions
│
├── 💼 commands/ - Command initialization
│
├── 💽 database/ - IO database management
│
├── 🚶 behavior/ - Mob pathfinding logic
│
├── 🧑‍🔧 modules/ - Player modules
│
└── 🌍 world/ - Initial reducers for world enhancement
```


src/modules/entity_behavior.js
```diff
@@ -25,8 +27,9 @@ export const trees = Object.fromEntries(
   }),
 )
 
+/** @type {import('../server').Module} */
```
DeltaEvo, line 30: I'm not sure that this works, last time that I tried that had no effect

src/modules/entity_target.js
```diff
@@ -21,9 +22,8 @@ export default {
     }
     return state
   },
-  /** @type {import('../context.js').Observer} */
   observe({ client, events, get_state }) {
-    events.on('MOB_ENTER_VIEW', ({ mob, signal }) => {
+    events.on('ENTITY_ENTER_VIEW', ({ mob, signal }) => {
```
DeltaEvo, line 26: Why was the name of this changed ?

I don't really mind, but if we do, we should do it consistently, so `reduce_mob` and all the `mob` variables as well.

I put mob originally because this was a subset of the entities in the minecraft sense, like for example traders are still entities but are not mobs

src/modules/entity_trader.js
```diff
@@ -52,23 +51,30 @@ function spawn_merchants({ client, events, world }) {
...
+        entityIds: traders_by_chunk.get(chunk_index(x, z)).map(({ id }) => id),
       })
     }
   })
 }
 
-/** @type {import('../context.js').Observer} */
 function open_trade({ client, world, get_state }) {
-  const { windowIds } = world.traders
+  const { traders_window_ids } = world
```
DeltaEvo, line 63: Is there a reason why you flattened this (and the others) ?

src/modules/entity_trader.js
```diff
@@ -99,15 +105,14 @@ function open_trade({ client, world, get_state }) {
   })
 }
 
-/** @type {import('../context.js').Observer} */
+// TODO: NPC should be handled through a behavior
```
DeltaEvo, line 108: This is by design, traders don't move and so don't need pathfinding and cross server synchronisation it's why those are separate.

We could however factor the code between the two because there is indeed some repetition

src/modules/player_chunk.js
```diff
@@ -0,0 +1,254 @@
...
+      unload_chunk({ client, x, z })
+      events.emit('CHUNK_UNLOADED', { x, z })
+    }
+
+    aiter(
+      abortable(
+        combine(
+          named_on(events, 'REQUEST_CHUNKS_LOAD', { signal }),
+          named_on(events, 'REQUEST_CHUNKS_UNLOAD', { signal }),
+          named_on(events, 'PROVIDE_CHUNKS', { signal }),
```
DeltaEvo, line 62: Why was this, split in two ?

I don't really understand the need.

src/modules/player_chunk.js
```diff
@@ -0,0 +1,254 @@
...
+        position,
+        view_distance,
+        teleport,
+      }))
+      .reduce(
+        async (last_state, state) => {
+          // this check has to be the first has it will initialize a player without chunks
+          // when the module is loaded its mandatory to send chunks in correct order
+          // previously it worked without it because we were loading modules
+          // before receiving the packet/settings from the client
```
DeltaEvo, line 151: You should not reference the history in comments, other than you and me, I think that no one knows what `previously` means.

src/core/modules.js
```diff
@@ -0,0 +1,35 @@
+const ADD_FUNCTIONS = [
+  'on',
+  'once',
+  'addListener',
+  'prependListener',
+  'prependOnceListener',
+]
+
+export function events_interceptor(emitter) {
```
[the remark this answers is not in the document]
DeltaEvo: Ok, I see why, you are putting multiple events_interceptor on one emitter, I don't really like this design and I feel that if we need to do that we should move away of EventEmitters.

By the way instead of using a proxy you can also use the `newListener` and `removeListener` event

src/modules/player_modules.js
```diff
@@ -0,0 +1,123 @@
...
+                    proxied_events.remove_all_intercepted_listeners()
+                    // and all local listeners that were added to the world
+                    proxied_world.remove_all_intercepted_listeners()
+                    // lastly we remove the listener from the global signal as we don't need it anymore
+                    signal.removeEventListener('abort', handle_global_abort)
+                  },
+                  { once: true },
+                )
+
+                await module.observe({
```
DeltaEvo, line 88: With this design this feels very wrong, this means that module loading is not atomic and that potentially we can miss events/state updates.

We should make observe sync or blocks action dispatch while loading modules for me.

src/core/modules.js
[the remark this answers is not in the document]
Sceat: moving away from it how ?

src/modules/entity_target.js
Sceat: Traders are technically entities and the way they are handled in their own module is quite old, they should have their own pathfinder looking at the player etc, we may want to make pnj walk etc during quests or simple lore

src/modules/entity_trader.js
Sceat: yes, ambiguity between arrays of entities and a subproperty managing them, especially world.mobs.all which sound absurd, also typings

src/modules/entity_trader.js
Sceat: see https://github.com/aresrpg/aresrpg/pull/696#discussion_r1313807435

src/modules/player_modules.js
Sceat: it won't be atomic anyway because each can be loaded/unloaded at any point in time (by design of not destroying everything to change the game state, so modules needs to be independant enough (that's why there is the `PROVIDE_MOBS` things and storing of ids

